PR #75, opened 2016-06-02
Hotfix/output

christiandompierre:
Also fixes some confusing help messages
Also remove the token string (-T) option

r?
@NCI-GDC/ucdevs


allisonheath:
@christiandompierre can you do a repeat of this PR to develop as well?

PR #76, opened 2016-06-16
fix(output, help): Explains what the prog bars are

christiandompierre:
Also fixes some confusing help messages
Also remove the token string (-T) option

in response to @allisonheath's request https://github.com/NCI-GDC/gdc-client/pull/75#issuecomment-226260023

r? @NCI-GDC/ucdevs


dankolbman:
lgtm

PR #89, opened 2016-08-22
chore(download): Add '.partial' extension

christiandompierre:
Adds a temporary extension to indicate file download still in progress. Goes with https://github.com/LabAdvComp/parcel/pull/85

Addresses DTT-1


dankolbman:
lgtm

PR #94, opened 2016-10-12
fix(args): Makes upload/download more consistent

christiandompierre:
Updated the way the upload accepts arguments to make it more
consistent with download
Allows upload (and deletion) or multiple files
Warns the user if they are using the deprecated -i flag
Created a new LogFormatter class to better format the logs
Now use different colors and format for different log level
Also changed default log level to WARNING
Also fixed some help messages

This addresses DTT-20 and DTT-11

### DIFF
--- a/gdc_client/upload/parser.py
+++ b/gdc_client/upload/parser.py
@@ -9,15 +9,21 @@ from . import manifest
 from . import exceptions
 
 from .client import GDCUploadClient
+from .. import log
 
+
+logger = log.get_logger('upload-client')
 
 def validate_args(parser, args):
     """ Validate argparse namespace.
     """
-    if args.manifest or args.identifier:
+    if args.identifier:
+        logger.warn('The use of the -i/--identifier flag has been deprecated.')
+
+    if args.manifest or args.file_ids:
         return
 
-    msg = 'must specify either --manifest or --identifier'
+    msg = 'must specify either --manifest or file_id'
     parser.error(msg)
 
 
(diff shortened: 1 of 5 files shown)


gdc_client/upload/parser.py
```diff
@@ -9,15 +9,20 @@
...
 
 
+logger = log.get_logger('upload-client')
+
 def validate_args(parser, args):
     """ Validate argparse namespace.
     """
-    if args.manifest or args.identifier:
+    if args.manifest or args.file_ids:
+        if args.identifier:
```
millerjs: hm, if a user specifies an `identifier` but not `manifest` or `file_ids`, then the deprecation warning won't be printed, but the `'must specify either --manifest or file_id'` will?

gdc_client/upload/parser.py
```diff
@@ -29,17 +34,18 @@ def upload(parser, args):
...
+                'project_id': args.project_id,
+                'path': args.path,
+                'upload_id': args.upload_id,
+            })
 
     # TODO remove debug - handled by logger
     debug = logging.DEBUG in args.log_levels
 
-    manifest_name = args.manifest.name if args.manifest else args.identifier
+    manifest_name = args.manifest.name if args.manifest else args.file_ids[0]
```
millerjs: does `nargs='*'` for `file_ids` allow 0 args? if so could `file_ids == []` causing an IndexError?

christiandompierre: No because it wont pass `validate_args`

gdc_client/upload/parser.py
millerjs: Yes, otherwise the user will not know they shouldn't be using that flag?

christiandompierre: Oh I think I see what you think and I believe you are confused about what `identifier` is. It's basically just a flag that gets set to true if the user specifies it. It's not a UUID. The DTT will exit with an error if `manifest` and `file_ids` are both missing even if `identifier` is specified. But I guess we can also warn the user about it's use of the `identifier` before exiting

millerjs: Ahh, yea, I think the user should be warned whenever they use a deprecated flag.  Still not sure I understand what `identifier` did, but will look later

christiandompierre: that was the old help message. This is no longer what shows when you print the help message. The `identifier` is kept only for backwards compatibility and is completely useless and ignored (except to print a warning).
I didn't print the warning if both `manifest` and `file_ids` were missing because I figured that the user has bigger problems to deal with if that's the case. I mean, I don't see a user specifying `-i` or `--identifier` without also specifying a `file_id` and if they do well the real problem is that they don't specify the `file_id`. The use of the `identifier` is deprecated but not an error per say.

In any case, I changed it to warn the user before exiting so they can fix that as well as fixing the missing `manifest` or `file_id`.

PR #96, opened 2016-10-12
fix(token): Makes token required

christiandompierre:
This addresses DTT-21

r? @NCI-GDC/ucdevs


christiandompierre:
the problem is that the token isn't required for download so I can't simply use `required=True` like I suppose you mean

christiandompierre:
redid PR from develop here: https://github.com/NCI-GDC/gdc-client/pull/99

PR #103, opened 2016-11-07
summary of retry changes

ProfOak:
comments and small cleanups

added retry functionality for failed downloads

added logging statements

@NCI-GDC/ucdevs r?

### DIFF
--- a/gdc_client/download/parser.py
+++ b/gdc_client/download/parser.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import time
 
 from functools import partial
 
@@ -12,6 +13,8 @@ from .. import log as logger
 from .client import GDCUDTDownloadClient
 from .client import GDCHTTPDownloadClient
 
+
+log = logging.getLogger('gdc-client')
 
 UDT_SUPPORT = ' '.join([
     'UDT is supported through the use of the Parcel UDT proxy.',
@@ -31,6 +34,7 @@ def validate_args(parser, args):
         parser.exit(status=1, message=UDT_SUPPORT)
 
 def get_client(args, token, **_kwargs):
+    # args get converted into kwargs
     kwargs = {
         'token': token,
         'n_procs': args.n_processes,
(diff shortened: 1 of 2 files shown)


gdc_client/download/parser.py
```diff
@@ -75,7 +82,48 @@ def download(parser, args):
         ids.add(i['id'])
 
     client = get_client(args, args.token_file)
-    client.download_files(ids)
+    downloaded_files, errors = client.download_files(ids)
+
+    if args.retry_amount > 0:
+        files_not_downloaded = []
+
+        for uuid, _ in errors.items():
```
dankolbman: same as `for uuid in errors.keys()`?

gdc_client/download/parser.py
```diff
@@ -75,7 +82,48 @@ def download(parser, args):
...
+
+
+def retry_download(client, uuid, retry_amount, no_auto_retry):
+
+    log.info('Retrying download {}'.format(uuid))
+
+    e = True
+    while 0 < retry_amount and e:
+        if no_auto_retry:
+            should_retry = raw_input('Retry download for {}? (y/N): '.format(uuid))
```
dankolbman: Don't know how I feel about accepting user input here. Can it just skip if the arg is set?

gdc_client/download/parser.py
ProfOak: Haha yeah I guess you're right. I think I was using the values before but scrapped it. Good eye.

gdc_client/download/parser.py
ProfOak: https://jira.opensciencedatacloud.org/browse/DTT-23 
This calls for a prompt. If you want to change that we might need more people's input on this.

dankolbman: oh ok, I see that you can still have it run automatically without retrying with `--retry-amount 0`.

gdc_client/download/parser.py
```diff
@@ -75,7 +82,48 @@ def download(parser, args):
...
+        else:
+            should_retry = 'y'
+
+        if should_retry.lower() == 'y':
+            log.info('{} retries remaning...'.format(retry_amount))
+            log.info('Retrying download... {} in 5 seconds'.format(uuid))
+            retry_amount -= 1
+            time.sleep(5)
```
madopal: Make the timeout a const

madopal:
I'm good assuming we make the wait for retry into a const or command line arg. I'm assuming we don't really know how long to wait, so it'd be good to be able to change it on the fly.

madopal:
Great, lgtm as well.

PR #107, opened 2016-11-18
Added file logging

ProfOak:
Rearranged code that wasn't using the get_logger func
modified imports to be more uniform

`--log-file LOG_FILE` can be used in conjunction with `--verbose` or `--debug` to affect the amount of details written to a file.

@NCI-GDC/ucdevs r?

It should also be known that this does not interact with any of the logging functionality that parcel handles. I'm not sure if that should be a consideration. Please advise.

### DIFF
--- a/gdc_client/log/log.py
+++ b/gdc_client/log/log.py
@@ -3,8 +3,9 @@ import sys
 
 from parcel import colored
 
-loggers = {}
 
+loggers = {}
+logger_filename = ''
 
 class LogFormatter(logging.Formatter):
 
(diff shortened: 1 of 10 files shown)


gdc_client/log/log.py
```diff
@@ -42,22 +43,33 @@ def format(self, record):
...
-    log.addHandler(handler)
-    loggers[name] = log
-    return log
+
+    stream_handler = logging.StreamHandler(sys.stderr)
+    stream_handler.setFormatter(formatter)
+
+    logger.addHandler(stream_handler)
+
+    if logger_filename and logger_filename != sys.stderr:
```
dankolbman: is this supposed to be `logger.logger_filename`?

dankolbman:
Was going to complain about the use of a global log, but I didn't realize how poorly laid out all the tools were. Might have to come back and refactor it all in the future.

gdc_client/log/log.py
ProfOak: It works correctly like this.

PR #109, opened 2016-11-21
fixed bug where the user would enter a full path, and resume would fail

ProfOak:
added a function to fix resume_path
added a test for the fix

Fixed a bug when a user would enter a path to a manifest.yml file.

https://github.com/NCI-GDC/gdc-client/blob/feat/python-26/gdc_client/upload/client.py#L148

This line just prepends 'resume_' to the path.

### DIFF
--- a/gdc_client/upload/client.py
+++ b/gdc_client/upload/client.py
@@ -120,6 +120,21 @@ def get_sleep_time(tries):
     return timeout * (0.5 + random.random()/2)
 
 
+def create_resume_path(file_path):
+    ''' in case the user enters a path, you want to create
+    a resume_filename.yml inside the same directory as the manifest.yml
+    '''
+
+    # check if it's a path or just a filename
+    if os.path.dirname(file_path):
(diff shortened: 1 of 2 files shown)


ProfOak:
@NCI-GDC/ucdevs r?

gdc_client/upload/client.py
```diff
@@ -120,6 +120,19 @@ def get_sleep_time(tries):
     return timeout * (0.5 + random.random()/2)
 
 
+def create_resume_path(file_path):
+    ''' in case the user enters a path, you want to create
+    a resume_filename.yml inside the same directory as the manifest.yml
+    '''
+
+    if '/' in file_path:
```
dankolbman: This doesn't work on windows? I think you can just remove this condition

ProfOak: There needs to be some distinction whether or not it's a path to a file or just a file, but I can definitely rephrase it another way to be more platform agnostic.

PR #110, opened 2016-11-21
fixed bug where the user would enter a full path, and resume would fail

ProfOak:
A user reported a bug that had an IOError where gdc-client tried to write in a directory that didn't exist.

According to this line of code, it just prepends "resume_" to a path.
https://github.com/NCI-GDC/gdc-client/blob/feat/python-26/gdc_client/upload/client.py#L148

So when that path starts with the root directory, it still prepends like "resume_" + /root/directory/manifest.yml

This fix makes it so the resume_manifest.yml is stored in the same path as the original manifest.

@NCI-GDC/ucdevs r?

### DIFF
--- a/gdc_client/upload/client.py
+++ b/gdc_client/upload/client.py
@@ -120,6 +120,21 @@ def get_sleep_time(tries):
     return timeout * (0.5 + random.random()/2)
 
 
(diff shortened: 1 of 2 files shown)


madopal:
:+1:

PR #112, opened 2016-12-06
fix(local_file_path): Make local_file_path do something

ProfOak:
local_file_path works when changed
replaced prints with logging

When uploading from a manifest file, the `local_file_path` key:value pair will be checked first. The backup will be to check the filename.

### DIFF
--- a/gdc_client/upload/client.py
+++ b/gdc_client/upload/client.py
@@ -16,7 +16,7 @@ from progressbar import ProgressBar, Percentage, Bar
 from collections import deque
 import time
 import copy
-from ..log import get_logger
+import logging
 
 from . import manifest
 
@@ -45,9 +45,7 @@ else:
     from mmap import ACCESS_READ
 
 
-log = get_logger('upload-client')
-log.propagate = False
-
+log = logging.getLogger('upload-client')
 
 def upload_multipart_wrapper(args):
     return upload_multipart(*args)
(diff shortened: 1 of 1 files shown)


ProfOak:
@NCI-GDC/ucdevs r?

gdc_client/upload/client.py
```diff
@@ -216,20 +216,24 @@ def get_files(self, action='download'):
                     self.file_entities.append(file_entity)
                     continue
 
-                path = f.get('path') or '.'
-                filename = f.get('file_name') or self.metadata['file_name']
-                file_entity.file_path = os.path.join(path, filename)
+                # first check local_file_path
+                # second check file_name
+                # https://github.com/NCI-GDC/gdcapi/pull/426#issue-146068652
+                file_entity.file_path = f.get('local_file_path') or \
```
christiandompierre: Please validate that `local_file_path` actually contains the file name and not only a path to the location of the file.

According to the variable name I am under the impression that you would have to append the file name to the path like this:
```
path = f.get('local_file_path') or '.'
filename = f.get('file_name') or self.metadata['file_name']
file_entity.file_path = os.path.join(path, filename)
```

gdc_client/upload/client.py
```diff
@@ -216,20 +216,29 @@ def get_files(self, action='download'):
...
-                path = f.get('path') or '.'
-                filename = f.get('file_name') or self.metadata['file_name']
-                file_entity.file_path = os.path.join(path, filename)
+                # first check local_file_path
+                # second check file_name
+                # https://github.com/NCI-GDC/gdcapi/pull/426#issue-146068652
+                file_entity.file_path = self.metadata['file_name']
+
+                # check to see if file exists in local_file_path
+                if not os.path.basename(f.get('local_file_path')) and os.path.exists(f.get_('local_file_path')):
```
christiandompierre: why `not`?

PR #121, opened 2017-06-30
Feat/dtt 45

ProfOak:
DTT has a exit code of 2 if a download has failed (1 was taken)


gdc_client/download/client.py
```diff
@@ -75,16 +75,27 @@ def parallel_download(self, stream, download_related_files=None,
...
 
     def __init__(self, uri, download_related_files=True,
                  download_annotations=True, *args, **kwargs):
-        # accepts args, but never called with args
+
         self.base_uri = self.fix_url(uri)
-        if not self.base_uri.endswith('/'):
-            self.base_uri += '/'
-        self.data_uri = urlparse.urljoin(self.base_uri, 'data/')
+        self.data_uri = urlparse.urljoin(self.base_uri, 'v0/data/')
```
madopal: Are you sure you want to hardcode to v0 here?

gdc_client/download/client.py
```diff
@@ -47,19 +47,19 @@ def download_small_groups(self, smalls):
                         filename = os.path.join(self.directory, filename.split('=')[1])
                     else:
                         filename = time.strftime("gdc-client-%Y%m%d-%H%M%S")
-                    log.info('Saving grouping {0}/{1}'.format(i+1, groupings_len))
+                    log.info('Saving grouping {}/{}'.format(i+1, groupings_len))
```
madopal: Don't think it's worth stopping for, but you're gonna hate having to go back & put those in for python 2.6.

ProfOak: Yeah I thought about that today, too. I'll probably put it all back and then some.

madopal: It's not necessary for this PR if you want to do it elsewhere, just sayin'. The question about hardcoding to one version below is more pressing, imho.

ProfOak:
Removed the explicit v0 @madopal

PR #127, opened 2017-10-27
Jbarno/feat/codacy cod cov

jbarno:
https://www.codacy.com/app/jbarno/gdc-client/dashboard

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,6 @@ install:
   - pip install -r dev-requirements.txt
   - python setup.py install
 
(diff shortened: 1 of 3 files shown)


ProfOak (approved):
lgtm

ProfOak:
Only thing I have to say is that if we start using this more then we should probably have some sort of CDIS account and pin everything to that.

jbarno:
Agreed to the cdis account. I wanted to get a POC out so we can see how devs feel about it. Seems like a lot of the issues are realllly nitpicky.

PR #128, opened 2017-10-27
Turn SSL on for metadata-query

jbarno:
Should probably be true :)


jbarno:
Here's the ticket: https://jira.opensciencedatacloud.org/browse/DTT-81

ProfOak (approved):

PR #135, opened 2018-03-22
Fix files not being added to download list

anmaxvl:
Update `separate_small_files` signature.

Update unit tests and method calls

### DIFF
--- a/gdc_client/query/index.py
+++ b/gdc_client/query/index.py
@@ -77,10 +77,10 @@ class GDCIndexClient(object):
         connections as possible.
 
         Args:
-            uuids (:obj: `list` of :obj: `str`): A list of UUIDs of the files
+            uuids (list): A list of UUIDs of the files
 
-        Returns:
-            :obj: `dict` of :obj: `dict`: metadata information
+        Return:
+            dict: metadata information
 
             self.metadata = {
                 str file_id: {
@@ -157,13 +157,13 @@ class GDCIndexClient(object):
         edge case.
 
         Args:
-            ids (:obj: `set` of :obj: `str`): a set of file UUIDs
+            ids (list): a set of file UUIDs
             chunk_size (int): the maximum allowed combined size of small files
 
         Return:
-            :obj: `list` of :obj: `str`: a list of big file UUIDs
-            :obj: `list` of :obj: `list` of :obj: `str`: a list of lists of
-                UUIDs. Each inner list representing a group of small file
+            list: a list of big file UUIDs
+            list: a list of lists of UUIDs. Each inner list representing a
+                group of small files
         """
 
         bigs = set()
--- a/gdc_client/query/index.py
+++ b/gdc_client/query/index.py
@@ -248,4 +248,4 @@ class GDCIndexClient(object):
 
         smalls = [ s for s in smalls if s != [] ]
 
-        return list(bigs), smalls
+        return bigs, smalls
--- a/gdc_client/query/index.py
+++ b/gdc_client/query/index.py
@@ -72,9 +72,15 @@ class GDCIndexClient(object):
         return json_response['data']['hits']
 
     def _get_metadata(self, uuids):
-        # type: List[str] -> Dict[str]str
-        """ Capture the metadata of all the UUIDs while making
-            as little open connections as possible.
+        """
+        Capture the metadata of all the UUIDs while making as little open
+        connections as possible.
+
+        Args:
+            uuids (:obj: `list` of :obj: `str`): A list of UUIDs of the files
+
+        Returns:
+            :obj: `dict` of :obj: `dict`: metadata information
 
             self.metadata = {
                 str file_id: {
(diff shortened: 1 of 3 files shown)


gdc_client/query/index.py
```diff
@@ -154,9 +155,18 @@ def separate_small_files(self,
...
+            chunk_size (int): the maximum allowed combined size of small files
+
+        Return:
+            list: a list of big file UUIDs
+            list: a list of lists of UUIDs. Each inner list representing a
+                group of small files
         """
 
-        bigs = []
+        bigs = set()
```
ProfOak, line 169: The resulting return is two lists, bigs and smalls. The reason that potential_smalls is a set is for an implicit uniqueness check. Please convert `bigs` to a list at the return of this function.

Bonus points: You don't have to update the docstring if you change it to a list. :-)

tests/test_query.py
```diff
@@ -111,34 +111,30 @@ def test_small_full_separate_small_files(self):
...
+                HTTP_CHUNK_SIZE)
 
         assert index.get_access('small') == uuids['small']['access']
         assert index.get_filesize('small') == uuids['small']['file_size']
         assert index.get_md5sum('small') == uuids['small']['md5sum']
         assert index.get_related_files('small') == uuids['small']['related_files']
         assert index.get_annotations('small') == uuids['small']['annotations']
 
-        assert bigs == ['small']
+        assert bigs == set(['small'])
```
ProfOak: Once bigs is a list, you can change this back.

tests/test_query.py
```diff
@@ -111,34 +111,30 @@ def test_small_full_separate_small_files(self):
...
+                HTTP_CHUNK_SIZE)
 
         assert index.get_access('small_no_friends') == uuids['small_no_friends']['access']
         assert index.get_filesize('small_no_friends') == uuids['small_no_friends']['file_size']
         assert index.get_md5sum('small_no_friends') == uuids['small_no_friends']['md5sum']
         assert index.get_related_files('small_no_friends') == []
         assert index.get_annotations('small_no_friends') == []
 
-        assert bigs == []
+        assert bigs == set()
```
ProfOak: this too

tests/test_query.py
```diff
@@ -151,17 +147,15 @@ def test_small_invalid_separate_small_files(self):
...
+                HTTP_CHUNK_SIZE)
 
         assert index.get_access(invalid) == None
         assert index.get_filesize(invalid) == None
         assert index.get_md5sum(invalid) == None
         assert index.get_related_files(invalid) == []
         assert index.get_annotations(invalid) == []
 
-        assert bigs == [invalid]
+        assert bigs == set([invalid])
```
ProfOak: this too

tests/test_query.py
```diff
@@ -170,27 +164,23 @@ def test_big_full_separate_small_files(self):
...
+                HTTP_CHUNK_SIZE)
 
         assert index.get_access('big') == uuids['big']['access']
         assert index.get_filesize('big') == uuids['big']['file_size']
         assert index.get_md5sum('big') == uuids['big']['md5sum']
         assert index.get_related_files('big') == uuids['big']['related_files']
         assert index.get_annotations('big') == uuids['big']['annotations']
 
-        assert bigs == ['big']
+        assert bigs == set(['big'])
```
ProfOak: this too

tests/test_query.py
```diff
@@ -206,16 +196,14 @@ def test_big_and_small_full_separate_small_files(self):
 
         # if a uuid has related files or annotations then they
         # are downloaded as big files
-        assert bigs == ['big', 'small']
+        assert bigs == set(['big', 'small'])
```
ProfOak: this too

tests/test_query.py
```diff
@@ -229,5 +217,5 @@ def test_big_and_small_no_rel_no_ann_separate_small_files(self):
         assert index.get_related_files('small_no_friends') == []
         assert index.get_annotations('small_no_friends') == []
 
-        assert bigs == ['big_no_friends']
+        assert bigs == set(['big_no_friends'])
```
ProfOak: this too

ProfOak (changes requested):

tests/test_query.py
anmaxvl: I will make it this way:
`assert set(bigs) == set(['big', 'small'])`
because there might be issue with the ordering of those items when converting from `set` to `list`

ProfOak: Set to list/list to set conversion preserves order. Doing this technically allows for information to be lost. Tests the return values as they are.

tests/test_query.py
```diff
@@ -206,16 +196,14 @@ def test_big_and_small_full_separate_small_files(self):
 
         # if a uuid has related files or annotations then they
         # are downloaded as big files
-        assert bigs == ['big', 'small']
+        assert set(bigs) == set(['big', 'small'])
```
ProfOak, line 199: don't type convert here

anmaxvl: The reason for this is, that the order might be messed up, even though the `bigs` might be `bigs == ['small', 'big']` this assertion won't work. Converting it to set, just makes sure that we don't need to worry about the order. Or do we?

ProfOak (approved):

PR #136, opened 2018-03-23
Make flags consistent. Minor clean up

anmaxvl:


gdc_client/defaults.py
```diff
@@ -33,4 +33,8 @@
 # proxy (on `proxy_host`) to traffic on the GDC api host
 proxy_port = 9000
 
-part_size = 1073741824 # 1GiB
+
+MB = 1024 * 1024
```
ProfOak: These are unused variables. This is probably fine how it is, otherwise you can just say `1024 * 1024 * 1024` or `1024**3`

gdc_client/download/parser.py
```diff
@@ -41,7 +41,6 @@ def get_client(args, index_client):
         'directory': args.dir,
         'segment_md5sums': args.segment_md5sums,
         'file_md5sum': args.file_md5sum,
-        'debug': logging.DEBUG in args.log_levels,
```
ProfOak, line 44: Why remove debug?

gdc_client/upload/parser.py
```diff
@@ -88,8 +87,8 @@ def config(parser):
     parser.add_argument('--server', '-s',
                         default=defaults.tcp_url,
                         help='GDC API server address')
-    parser.add_argument('--part-size', '-ps',
-                        default=defaults.part_size,
+    parser.add_argument('--http-chunk-size', '-c',
```
ProfOak, line 90: Make sure this is clearly documented in the ticket, so user services knows to relay this information to the users.

ProfOak (changes requested):

gdc_client/download/parser.py
anmaxvl: I think these flags are handled by the logger.

gdc_client/upload/parser.py
anmaxvl: got it

gdc_client/defaults.py
anmaxvl: alright

ProfOak (approved):

PR #138, opened 2018-03-28
Fix bug with wrong kwargs. Bump version

anmaxvl:


gdc_client/version.py
```diff
@@ -1 +1 @@
-__version__ = 'v1.3.0'
+__version__ = 'v1.3.1'
```
ProfOak: This is probably ok for now but we don't normally bump version numbers until we're ready to do a release.

gdc_client/defaults.py
```diff
@@ -34,3 +34,6 @@
 proxy_port = 9000
 
 HTTP_CHUNK_SIZE = 1024 * 1024 * 1024  # 1 GiB
+
+# Bringing this back to avoid errors in other modules
+part_size = HTTP_CHUNK_SIZE
```
ProfOak: Where else uses this?

gdc_client/version.py
anmaxvl: Ah ok. This is something that Ian asked me to do for his build. Will put it back.

gdc_client/defaults.py
anmaxvl: Just double checked, seems to be nowhere else. But `repl` was still using it and I already changed it to `HTTP_CHUNK_SIZE`

ProfOak (approved):

PR #146, opened 2018-08-06
fix: annotation download when active api fail try legacy api

khan08:
FIX: For legacy file with annotation, gdc-client can't find annotation.
CAUSE: gdc-client only calls active api for downloading annotation, it should also try calling legacy api if active api throws error.

This solution is a quick inline fix.


gdc_client/download/client.py
```diff
@@ -65,12 +65,19 @@ def download_annotations(self, file_id):
...
         if annotations:
-            log.debug('Found {0} annotations for {1}.'.format(
-                len(annotations), file_id))
+            log.debug('Found {0} annotations for {1}.'.format(len(annotations), file_id))
+            # TODO: Refactor this with the _post method for a generic function for calling active/legacy api
             r = requests.get(
                 urlparse.urljoin(self.data_uri, annotation_list),
                 params={'compress': True},
                 verify=self.verify)
+            if r.status_code != requests.codes.ok:
```
anmaxvl: not sure how feasible it's here, but we added a `203` response code for files that have newer versions. Maybe stick the check here as well, just in case.

anmaxvl (changes requested):

gdc_client/download/client.py
```diff
@@ -65,12 +65,19 @@ def download_annotations(self, file_id):
...
+                # try legacy if active doesn't return OK
+                r = requests.get(
+                    urlparse.urljoin(self.legacy_data_uri, annotation_list),
```
ProfOak: This might go over the max character limit of a URL. You can hit this endpoint with a POST request and give it a body containing the IDs.

ProfOak (changes requested):

ProfOak (approved):

anmaxvl (approved):
lgtm!

PR #147, opened 2018-08-14
[feat][DTT-112]: Add --latest flag to download latest versions of files.

anmaxvl:


gdc_client/download/client.py
```diff
@@ -94,13 +96,17 @@ def _untar_file(self, tarfile_name):
...
         return [ m.name for m in members ]
 
-
     def _md5_members(self, members):
         # type: (List[str]) -> List[str]
         """ Calculate md5 hash and compare them with values given by the API """
 
         errors = []
         for m in members:
+            if re.findall(re.compile(SUPERSEDED_INFO_FILENAME_TEMPLATE), m):
```
kulgan: You can call re.compile(SUPERSEDED_INFO_FILENAME_TEMPLATE) once and reuse it, since its basically the going to remain the same all through. Or maybe change SUPERSEDED_INFO_FILENAME_TEMPLATE to return the compiled regular expression object

gdc_client/utils.py
```diff
@@ -0,0 +1,14 @@
+from urllib import urlencode
+
+
+def url_with_params(path, *params, **kwparams):
```
kulgan: Can you rename this to something like `build_url` or something more in the line of what the code does, `url_with_params` was a unclear when I first saw it. Also, can you add unit tests for it and some docstring maybe

gdc_client/utils.py
```diff
@@ -0,0 +1,14 @@
+from urllib import urlencode
```
ProfOak: The `requests` library actually supports this out of the box.

```python
>>> payload = {'key1': 'value1', 'key2': 'value2'}
>>> r = requests.get('http://httpbin.org/get', params=payload)
>>> print(r.url)
http://httpbin.org/get?key2=value2&key1=value1
```

http://docs.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls

gdc_client/download/parser.py
```diff
@@ -123,10 +127,13 @@ def download(parser, args):
         log.debug('Downloading big files...')
 
         # create URLs to send to parcel for download
-        bigs = [ urlparse.urljoin(client.data_uri, b) for b in bigs ]
-        downloaded_files, big_error_dict = client.download_files(bigs)
+        params = ('latest',) if args.latest else ()
+        bigs = [
+            urlparse.urljoin(client.data_uri, url_with_params(b, *params))
```
ProfOak: Be careful with urljoin. It does some URL validation. I think this should be OK if we are re-formatting urls as we get them through the `--server` flag.
```python
In [1]: from urlparse import urljoin

In [2]: urljoin('a.b.c', 'data')
Out[2]: 'data'

In [3]: urljoin('https://a.b.c', 'data')
Out[3]: 'https://a.b.c/data'
```

kulgan (changes requested):

gdc_client/defaults.py
```diff
@@ -34,3 +34,6 @@
 proxy_port = 9000
 
 HTTP_CHUNK_SIZE = 1024 * 1024 * 1024  # 1 GiB
+
+# The following file will contain superseded files information
+SUPERSEDED_INFO_FILENAME_TEMPLATE = 'superseded_files[0-9._]+.txt'
```
ProfOak: `\d` is shorthand for `[0-9]`

gdc_client/download/client.py
```diff
@@ -136,7 +141,7 @@ def _post(self, path, headers={}, json={}, stream=True):
                 json=json,
                 headers=headers,
             )
-            if r.status_code != requests.codes.ok:
+            if r.status_code not in [requests.codes.ok, 203]:
```
ProfOak: If you want, for consistency, you can change `requests.codes.ok` to `200`. Pylint also complains about it because of how requests is set up, so I don't feel bad about changing it. :p

gdc_client/download/client.py
```diff
@@ -180,7 +186,7 @@ def _download_tarfile(self, small_files):
             log.error(r.text)
             return '', []
 
-        if r.status_code != requests.codes.ok:
+        if r.status_code not in [requests.codes.ok, 203]:
```
ProfOak: Same as comment above.

gdc_client/utils.py
anmaxvl: right, the idea is that it won't support stuff like `url?flag1&flag2` and that's how `latest` and some other flags (`tar`?) are passed

gdc_client/download/client.py
anmaxvl: right.

gdc_client/utils.py
anmaxvl: will do

gdc_client/download/client.py
anmaxvl: ok 😄

gdc_client/download/parser.py
anmaxvl: I think we do `fix_url` stuff, which _should_ handle this.

gdc_client/utils.py
jbarno: will break in python3
```
if sys.version_info >= (3,):
    import urllib.parse as urlparse
    from urllib.parse import urlencode
else:
    import urlparse
    from urllib import urlencode

```

gdc_client/utils.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def url_with_params(path, *params, **kwparams):
+    final_params = []
+    if params:
+        final_params.append('&'.join(params))
```
jbarno: Isn't the ampersand taken care of in the return statement? I think this might make && depending on how it is called.

jbarno (changes requested):
py3 fun. Also I think the url_with_params has a bug

gdc_client/utils.py
ProfOak:
```python
In [1]: import requests

In [2]: requests.post('https://api.gdc.cancer.gov/status', params='a&b&c').url
Out[2]: u'https://api.gdc.cancer.gov/status?a&b&c'
```
The `params` keyword arg accepts any string or dictionary.

gdc_client/defaults.py
anmaxvl: 👍

gdc_client/utils.py
anmaxvl: There are 2 parts here. Regular params will look like `a&b&c` and key word params like `d=e&f=g`, so the `final_params == ['a&b&c', 'd=e&f=g']`. E.g.

    >>> from gdc_client.utils import url_with_params
    >>> url_with_params('http://example/com/some/path', 'latest', 'tarfile', foo='bar', baz='foo')
    'http://example/com/some/path?latest&tarfile&foo=bar&baz=foo'

gdc_client/utils.py
anmaxvl: this isn't the only thing that's going to break with python3 😄

requirements.txt
```diff
@@ -1,9 +1,9 @@
 cmd2==0.6.8
-cryptography==2.2.1
+cryptography==2.3.1
```
jbarno, line 2: danke. There's a bot that should do it too!

jbarno (approved):

kulgan (approved):

PR #148, opened 2018-08-17
[feat][DTT-105][DTT-30] : Remove interactive mode

anmaxvl:
* remove interactive mode
* add settings sub-command to display default settings
* refactor code: flags store what they are supposed to, change initializations to work with refactored flags
* add `--config` flag to provide support for custom `INI`-type configuration files for `upload` and `download`, also load a user-config located at `~/.config.dtt`
* If config is passed to `settings` sub command, it will show how the `gdc-client` configuration will look like using the specified config file.


gdc_client/common/config.py
```diff
@@ -0,0 +1,92 @@
...
+            option: self.get_setting(section, option)
+            for section in self.sections for option in self.setting_getters
+            if self.get_setting(section, option) is not None
+        }
+
+    def get_setting(self, section, option):
+        try:
+            getter = getattr(self.config, self.setting_getters[option])
+            return getter(section, option)
+        except (ConfigParser.NoOptionError, ConfigParser.NoSectionError,
```
ProfOak: I would add a debug log here. It looks like it should continue to run with whatever the default value is, even if there is an error in the config file. It won't be obvious to the user or the support team if a user has problems with their config file. This could also be an opportunity to break up the exception clauses based on what you want to log.

gdc_client/upload/parser.py
```diff
@@ -71,31 +70,31 @@ def config(parser):
...
+    conf = GDCClientUploadConfig()
+    upload_defaults = conf.to_dict()
+    upload_defaults['func'] = func
+
+    parser.set_defaults(**upload_defaults)
 
     parser.add_argument('--project-id', '-p', type=str,
                         help='The project ID that owns the file')
-    parser.add_argument('--path', '-f', metavar='path', default='',
+    parser.add_argument('--path', '-f', metavar='path', type=str,
```
ProfOak, line 78: The defaults are going to come from the new config file right? That's neat.

ProfOak (changes requested):

gdc_client/upload/parser.py
anmaxvl: yes, this was the idea, but it seems like the config gets lost during packaging... need to figure out another way of loading the default one... might end up storing it in a dictionary it seems...

anmaxvl: ended up hard coding the defaults in code, because it's wonky to ensure that a configuration file is packaged properly.

anmaxvl:
@ProfOak, please, revisit. Should be good now.

gdc_client/settings/parser.py
```diff
@@ -0,0 +1,42 @@
...
+
+class SettingsResolver(object):
+    def __init__(self):
+        self.config = GDCClientConfigShared()
+
+    def ls(self, section, args):
+        log.info(self.config.to_display_string(section))
+        return self.config.to_display_string(section)
+
+    def list(self, section, args):
```
ProfOak: I get that the function name is linked to the config option, but is there any way to make a special case for this `list` command? I feel funny whenever I see python keywords get clobbered.

ProfOak (approved):
The `list` thing isn't a big deal. :+1:

gdc_client/settings/parser.py
anmaxvl: yeah, maybe `show`? 😄

anmaxvl: or maybe just leave `ls`.

gdc_client/upload/parser.py
```diff
@@ -67,35 +65,32 @@ def upload(parser, args):
...
                         help='directory path to find file')
     parser.add_argument('--upload-id', '-u',
                         help='Multipart upload id')
-
     parser.add_argument('--insecure', '-k',
-                        action='store_false',
```
jbarno, line 84: Does this disable SSL within requests?

requirements.txt
```diff
@@ -1,4 +1,3 @@
-cmd2==0.6.8
```
jbarno, line 1: Great!

jbarno (commented):
https://github.com/NCI-GDC/gdc-client/pull/148/files#diff-442281cfa7283e381b9a22ec27771700R83

Please turn SSL back on

PR #158, opened 2018-10-04
[feat][DTT-30]: Feat/dtt 30 feedback from ian

anmaxvl:
* Remove parent parsers for `settings` sub parser (no need for `--token` etc). Refactor settings
* Change the way settings are displayed. Now the output matches an `INI`-type config file
* Override `ArgumentParser.error` to display full help message in addition to default error message

### DIFF
--- a/gdc_client/common/config.py
+++ b/gdc_client/common/config.py
@@ -5,7 +5,11 @@ from gdc_client.defaults import (
     processes, USER_DEFAULT_CONFIG_LOCATION, HTTP_CHUNK_SIZE
 )
 
-log = logging.getLogger('gdc-client-config')
+log = logging.getLogger('gdc-client')
+
+# This will display the default configs in a INI-type format, so that users
+# will be able to copy and modify as needed
+DISPLAY_TEMPLATE = '[{}]\n{}\n'
 
 
 class GDCClientConfigShared(object):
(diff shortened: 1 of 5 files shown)


anmaxvl:
Reopened PR so that JIRA can pick it up correctly.

ProfOak:
What was the purpose of closing the pr? Do you need a review still?

anmaxvl:
@ProfOak, it wasn't linking to the JIRA ticket, because of the wrong title. I updated it and reopened. Yes, please, review. Thanks!

gdc_client/common/config.py
```diff
@@ -1,11 +1,28 @@
...
+
+
+class GDCClientArgumentParser(argparse.ArgumentParser):
+    """This is a workaround introduced here https://groups.google.com/forum/#!topic/argparse-users/LazV_tEQvQw
+    which enables to print the full help message in case something went wrong
+    with argument parsing
+    """
+    def error(self, message):
+        self.print_help(sys.stderr)
+        sys.stderr.write('\ngdc-client error: {}\n'.format(message))
```
ProfOak, line 24: Why stderr instead of logging?

ProfOak: I only ask because I think the dtt-ui pays attention to this sort of thing.

https://github.com/NCI-GDC/dtt-ui/search?q=stdout&unscoped_q=stdout

anmaxvl: Oh, I see. I will take a look. I was afraid that logger might not be setup at this point.

ProfOak: Oh I didn't consider that. This is probably ok then.

ProfOak (approved):

PR #161, opened 2018-10-19
fix(dtt-121): add pre-check for latest files

ProfOak:
The problem was that the directory created was done so using the UUID in
the url. Then we were doing `/v0/data/uuid?latest`, so the directory
name was the always the old UUID even if there was a new version.

Using the /files/versions endpoint we can tell the latest version of a
file before going to download it. Replacing the old list of UUIDs with
the new list of UUIDs with the newer versions in it. After that the download
process continues as normal.

### DIFF
--- a/gdc_client/query/versions.py
+++ b/gdc_client/query/versions.py
@@ -1,0 +1,55 @@
+"""gdc_client.query.versions
+
+Functionality related to versioning.
+"""
+import requests
+
+
+def get_latest_versions(url, uuids):
+    """Get the latest version of a UUID according to the api.
+
+    Args:
+        url (str):
+        uuids (list): list of UUIDs that might have a new version
+
+    Returns:
+        list: list of uuids with potentially new versions
+    """
+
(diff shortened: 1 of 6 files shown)


gdc_client/query/versions.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Args:
+        elements (list): list of elements
+        chunk_size (int): max number of elements in the resulting chunk
+
+    Returns:
+        list: list of lists containing chunked workloads
+    """
+
+    chunks = [[]]
```
anmaxvl: maybe replace this with a generator?

    current_index = 0
    while current_index < len(elements):
        yield elements[current_index:current_index + chunk_size)
        current_index += chunk_size

anmaxvl (dismissed):
Neat workaround!

anmaxvl (commented):
@kulgan, please review

kulgan (approved):
lgtm

PR #163, opened 2018-10-25
fix(segment md5sum): allows disabling and enabling md5 validation of download segments

kulgan:


ProfOak:
Not that it matters, but the pep for type hints say to use caps on non-primative types.

ProfOak:
Not that it matters too much but the type hints pep says to use caps for types that aren't technically primitive types in other languages. You can see it in this example:

```python
def send_email(address,     # type: Union[str, List[str]]
               sender,      # type: str
               cc,          # type: Optional[List[str]]
               bcc,         # type: Optional[List[str]]
               subject='',
               body=None    # type: List[str]
               ):
    # type: (...) -> bool
    """Send an email message.  Return True if successful."""
    <code>
```

It mostly has to do with the fact that these types aren't part of the language, but come in a new `typing` module. They are defined as classes (CapitalizedNames) and you have to import them. Here's a list of what all the new types are, and how to make your own in python 3.6+

https://docs.python.org/3/library/typing.html

I had started putting type hints in an effort to improve documentation but I'd like to kind of want to get rid of them in favor of Google style docstrings. Just because that's what we use everywhere else now.

https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code

gdc_client/download/client.py
```diff
@@ -19,10 +19,49 @@
...
+    if not url.endswith('/'):
+        url = '{0}/'.format(url)
+
+    if not (url.startswith('https://') or url.startswith('http://')):
+        url = 'https://{0}'.format(url)
+
+    return url
+
+
+class GDCHTTPDownloadClient(Client):
```
ProfOak: This is a lot more work than necessary. I started a cleanup project but got side-tracked. Here's how I started doing it with the mixin class and multiple inheritance. If you don't define an `__init__()` method then the superclass' method gets called.

https://github.com/NCI-GDC/gdc-client/blob/chore/code-style-cleanups/gdc_client/download/client.py

kulgan: Yea, I kinda noticed that but this is what actually pushed me to implement it like this https://github.com/NCI-GDC/gdc-client/blob/5f59d8502c4b4fe6c7a6709f4641bf980548bd91/gdc_client/download/client.py#L291

Which is intended to override the `parallel_download` method in the parcel `Client` class. This is an unusual way of doing inheritance, the `HTTPClient` also adds no new functionality, so can be removed and inherit directly from the `Client`. In a way we are basically on the same track, but keeping the `GDCUDTClient`

gdc_client/download/client.py
```diff
@@ -9,7 +9,7 @@
 from StringIO import StringIO
 
 import requests
-from parcel import HTTPClient, UDTClient, utils
+from parcel import Client, UDTClient, utils
```
ProfOak: Can you change this back to HTTPClient? I know it's just an alias for Client in parcel but everything in this repository is named "client" and this change does not help readability.

gdc_client/download/client.py
```diff
@@ -19,10 +19,49 @@
 log = logging.getLogger('gdc-download')
 
 
-class GDCDownloadMixin(object):
+def fix_url(url):
```
ProfOak, line 22: There was no self use so it could be broken out into it's own function, nice. :clap:

gdc_client/query/index.py
```diff
@@ -125,9 +126,9 @@ def _get_metadata(self, uuids):
 
         for h in active_hits + legacy_hits:
             related_returns = h.get('index_files', []) + h.get('metadata_files', [])
-            related_files = [ r['file_id'] for r in related_returns ]
+            related_files = [r['file_id'] for r in related_returns ]
```
ProfOak: If you're cleaning up whitespace, take it off the right-hand-side too.

gdc_client/download/client.py
```diff
@@ -19,10 +19,49 @@
...
+
+        self.base_uri = uri
+        self.data_uri = urlparse.urljoin(self.base_uri, 'data/')
+
+        self.annotations = download_annotations
+        self.related_files = download_related_files
+
+        self.md5_check = kwargs.get('file_md5sum')
+
+        self.gdc_index_client = index_client
```
ProfOak, line 60: Thank you, this "index_client" might have been confusing now that indexd is in use.

ProfOak (changes requested):

ProfOak (approved):

PR #164, opened 2018-10-26
[chore][DTT-105]: rename flags. change defaults

anmaxvl:
After reviewing the code and discussing with Ian, we came to a conclusion that `http_chunk_size` serve different purposes for upload and download. To accommodate that we rename `--http-chunk-size` flag for upload to `--upload-part-size` and add corresponding default values in `defaults.py`.

### DIFF
--- a/gdc_client/upload/client.py
+++ b/gdc_client/upload/client.py
@@ -69,7 +69,7 @@ class Stream(object):
         return self._file.read(num)
 
 
-def upload_multipart(filename, offset, bytes, url, upload_id, part_number,
+def upload_multipart(filename, offset, num_bytes, url, upload_id, part_number,
                      headers, verify=True, pbar=None, ns=None):
     tries = MAX_RETRIES
     while tries > 0:
@@ -79,14 +79,14 @@ def upload_multipart(filename, offset, bytes, url, upload_id, part_number,
             if OS_WINDOWS:
                 chunk_file = mmap(
                     fileno=f.fileno(),
-                    length=bytes,
+                    length=num_bytes,
                     offset=offset,
                     access=ACCESS_READ
                 )
             else:
                 chunk_file = mmap(
                     fileno=f.fileno(),
-                    length=bytes,
+                    length=num_bytes,
                     offset=offset,
                     prot=PROT_READ
                 )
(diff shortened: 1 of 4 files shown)


gdc_client/upload/parser.py
```diff
@@ -85,7 +82,7 @@ def config(parser, upload_defaults):
     # TODO remove this and replace w/ top level host and port
     parser.add_argument('--server', '-s', type=str,
                         help='GDC API server address')
-    parser.add_argument('--http-chunk-size', '-c', type=int,
+    parser.add_argument('--upload-part-size', '-c', type=int,
```
kulgan, line 91: Leaving the `--http-chunk-size` and returning a message that says no longer supported, use `--upload-part-size` for same result might help users who already have anything that uses `--http-chunk-size` for upload

gdc_client/upload/client.py
```diff
@@ -461,15 +463,15 @@ def upload_parts(self):
...
         self.total_parts = part_amount
         self.pbar = ProgressBar(
             widgets=[Percentage(), Bar()], maxval=self.total_parts).start()
         try:
             for i in xrange(part_amount):
-                offset = i * self.part_size
+                offset = i * self.upload_part_size
                 remaining_bytes = self.file_size - offset
-                bytes = min(remaining_bytes, self.part_size)
+                bytes = min(remaining_bytes, self.upload_part_size)
```
kulgan: You might want to rename this variable, its same as built in type `bytes`

gdc_client/upload/client.py
```diff
@@ -139,7 +139,7 @@ def create_resume_path(file_path):
 
 class GDCUploadClient(object):
 
-    def __init__(self, token, processes, server, part_size,
+    def __init__(self, token, processes, server, upload_part_size,
```
kulgan, line 142: using empty dict as default parameter for `files`

kulgan (changes requested):

gdc_client/upload/parser.py
anmaxvl: It was `--part-size` before, but I see your point.

gdc_client/upload/client.py
anmaxvl: sure

gdc_client/upload/client.py
anmaxvl: yeaah....

gdc_client/upload/parser.py
anmaxvl: After discussing with Ian, we decided to have `--part-size`, but mark it as deprecated and remove with the next release.

anmaxvl:
@kulgan, let me know if this is good.

kulgan (approved):

PR #174, opened 2018-11-15
[fix][DTT-126]: fix(legacy): download of legacy

anmaxvl:


gdc_client/query/versions.py
```diff
@@ -31,13 +31,13 @@ def get_latest_versions(url, uuids):
...
         for result in resp.json():
-            # Probably wouldn't happen, but just in case add an error check.
             file_id = result.get("id")
             uuid = result.get('latest_id')
             if uuid:
                 latest_versions[file_id] = uuid
                 continue
-            logger.info("No latest version found for {}".format(file_id))
+            # Might happen for legacy files
+            latest_versions[file_id] = file_id
```
ProfOak: Tthis overwrites the previous block's results. You need to extend the previous block with an else case.

```python
if uuid:
    latest_versions[file_id] = uuid
else:
    latest_versions[file_id] = file_id
```

ProfOak (changes requested):

gdc_client/query/versions.py
anmaxvl: it won't, since there is a `continue` statement.

anmaxvl: I added unit tests for it last time as well.

ProfOak: Oh I see, sorry I misread. You could rewrite this without a continue though.

anmaxvl: yeah, sure

ProfOak (approved):
:+1:

PR #175, opened 2018-11-19
feat(dtt-125): add ssl verification on query client

ProfOak:


jbarno (approved):
I do not not like it. 

lgtm

PR #178, opened 2018-11-26
[fix][DTT-128]: force compression when downloading annotations

anmaxvl:


jbarno (approved):
lgtm

jbarno:
Any other time we need to pass compress flag? Like related files or previous versions?

anmaxvl:
@jbarno, not that I am aware of. It seems to be the only place where we expect a tar file, but the API might not return one.

ProfOak:
The original goal was to never use the compress flag as it puts extra load on the api servers. I suppose annotations are small enough where it probably wouldn't be an issue.

anmaxvl:
@ProfOak are you ok with this? Other solution would be to wrap it inside `try except` and change corresponding logic.

ProfOak:
Yes this is fine by me. :-p

I don't think annotations would be a problem as they are just a few lines of text. That's a lot less strenuous than trying to compress several multi-gigabyte files. Even then that could be seen as a good thing as long as the files can be compressed well enough. Compression might make it so the end-user wouldn't have to download as much. I'm not sure where the most optimal path is currently.

PR #247, opened 2019-08-16
TT-1093 set up tox for gdc-client

dariatarasova:
Not finished
in Python 2, there is one test failing (test_download_tarfile)
in Python 3, there is a moduleNotFoundError: no module named server pertaining to the parcel module. I believe other modules simply had to import another library in order to deal with this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,8 @@ python:
 sudo: false
 
 install:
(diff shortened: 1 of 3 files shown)


anmaxvl:
has been addressed by https://github.com/NCI-GDC/gdc-client/pull/248

PR #248, opened 2019-09-16
[FEAT] TT-1138 Remove CDIS Dependencies

huarngpa:
* Move parcel over
* Add tox
* Make Python3 compatible


gdc_client/parcel/client.py
```diff
@@ -0,0 +1,253 @@
+from . import const
```
anmaxvl: how copy-paste is this file? if most of it is, please, include the link or reference to the original author.

gdc_client/parcel/defaults.py
```diff
@@ -0,0 +1,9 @@
+tcp_url = 'https://localhost'
+udt_url = 'https://localhost'
```
anmaxvl: 💣

tests/test_query.py
```diff
@@ -32,31 +32,31 @@ def tearDown(self):
     def test_no_metadata_get_related_files(self):
         index = GDCIndexClient(uri=base_url)
 
-        results = index.get_related_files(uuids['small'])
+        results = index.get_related_files('small')
```
ProfOak, line 35: Can you elaborate on this change? I think this method accepts a list of uuids.

gdc_client/download/client.py
```diff
@@ -320,24 +326,3 @@ def parallel_download(self, stream):
                     file_id, e))
                 if self.debug:
                     raise
-
-
-class GDCUDTDownloadClient(GDCHTTPDownloadClient):
```
ProfOak, line 325: This is probably ok. I don't think there are plans to use UDT in the near future.

ProfOak (commented):
Seems reasonable but I had one question.

gdc_client/download/client.py
huarngpa: Yeah, just remove it :(

tests/test_query.py
huarngpa: `uuids['small']` is a [dict](https://github.com/NCI-GDC/gdc-client/blob/a7f147c5ad4a60dc68784de7e3a7cc48f3f2bde6/tests/conftest.py#L46). I think in python2 this was fine (possibly dict types could be dictionary keys)?? But python3 doesn't allow this. See this [line](https://github.com/NCI-GDC/gdc-client/blob/a7f147c5ad4a60dc68784de7e3a7cc48f3f2bde6/gdc_client/query/index.py#L19) for what I mean... it expects a type string and then searches through a list of dictkeys

ProfOak: Ah that makes sense. My mistake.

ProfOak (approved):

gdc_client/parcel/download_stream.py
```diff
@@ -0,0 +1,290 @@
+from . import utils
```
anmaxvl: same here

gdc_client/parcel/client.py
```diff
@@ -0,0 +1,253 @@
...
+                        raise
+                    else:
+                        log.error("Download aborted: {0}".format(str(e)))
+                        break
+
+        # Divide work amongst process pool
+        pool = [Process(target=download_worker) for i in range(n_procs)]
+
+        # Start pool
+        map(lambda p: p.start(), pool)
```
anmaxvl: not gonna work in py3. and a pretty bad practice in general to use the side-effect behavior 🤷‍♂

gdc_client/parcel/log.py
```diff
@@ -0,0 +1,28 @@
+import logging
+import sys
+from .portability import colored
+
+
+loggers = {}
+
+
+# Logging
+def get_logger(name='parcel'):
```
anmaxvl, line 19: looks familiar 🤔 :trollface:

gdc_client/parcel/segment.py
```diff
@@ -0,0 +1,280 @@
...
+        self.pbar = None
+        self.pbar = get_pbar(self.download.url, self.download.size)
+
+    def _setup_work(self):
+        if self.is_complete():
+            log.debug('File already complete.')
+            return
+
+        work_size = self.integrate(self.work_pool)
+        self.block_size = work_size / self.n_procs
```
anmaxvl: `//` to avoid float

gdc_client/parcel/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+def check_transfer_size(actual, expected):
+    """Simple validation on any expected versus actual sizes.
+
+    :param int actual: The size that was actually transferred
+    :param int actual: The size that was expected to be transferred
+
+    """
+
+    return actual == expected
```
anmaxvl, line 39: hm, ok 🤷‍♂

gdc_client/parcel/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        s=' {0} '.format(file_id), n=50, c='-')))
+
+
+def print_closing_header(file_id):
+    log.debug('^{0}^'.format('{s:{c}^{n}}'.format(
+        s=' {0} '.format(file_id), n=50, c='-')))
+
+def write_offset(path, data, offset):
+    try:
+        f = open(path, 'r+b')
```
anmaxvl: mb do `with open(path, 'r+b') as f:` instead?

gdc_client/parcel/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        f.seek(offset)
+        f.write(data)
+        f.close()
+    except Exception as e:
+        raise Exception('Unable to write offset: {0}'.format(str(e)))
+
+
+def read_offset(path, offset, size):
+    try:
+        f = open(path, 'r+b')
```
anmaxvl: same as above

gdc_client/parcel/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        return data
+    except Exception as e:
+        raise Exception('Unable to read offset: {0}'.format(str(e)))
+
+
+def set_file_length(path, length):
+    try:
+        if os.path.isfile(path) and os.path.getsize(path) == length:
+            return
+        f = open(path, 'wb')
```
anmaxvl: same

anmaxvl (changes requested):

gdc_client/parcel/defaults.py
huarngpa: gone

gdc_client/parcel/client.py
huarngpa: good point!

gdc_client/parcel/log.py
huarngpa: lol

gdc_client/parcel/utils.py
huarngpa: lol.. let's just move on

gdc_client/parcel/utils.py
huarngpa: good point

gdc_client/parcel/client.py
huarngpa: thanks for the catch!

anmaxvl (approved):
lgtm

gdc_client/download/client.py
```diff
@@ -5,12 +5,17 @@
 import sys
 import tarfile
 import time
-import urlparse
-from StringIO import StringIO
+from six.moves.urllib.parse import urljoin
```
kulgan: Can you do an as instead
``` python
from six.moves.urllib.parse as urlparse
```

gdc_client/parcel/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .http_client import HTTPClient
```
kulgan: Can you avoid relative imports

gdc_client/download/client.py
huarngpa: ah, clever :+1:

gdc_client/parcel/__init__.py
huarngpa: Fixing

PR #249, opened 2019-12-05
feat(package): upgrade package script to build from py3

ProfOak:
- Include instructions in readme on how to use package script, for
building a PyInstaller single executable.
- Put requests back in requirements.txt because it wasn't being
installed until dev-requirements.txt was installed. (Not needed for
"production" builds)
- Bump some deps to get wheels for later versions of python on Windows.
- Windows specific changes to package script.
- Bump version number to next minor version.

With the help of @bilalbaqar we were able to build binaries for Windows, Linux, and Mac OS.


bin/package
```diff
@@ -6,7 +6,7 @@ function cleanup() {
     # get rid of the virtualenv
     echo "Cleaning up"
     deactivate
-    rm -rf $VENV_NAME/ build/ dist/ include/ local/ share/  gdc-client.spec
```
singergr-ctds, line 9: :eyes:

bin/package
```diff
@@ -76,12 +78,12 @@ cd bin
 echo "Building for ${TARGET_ENVIRONMENT}..."
 
 # Make sure the correct version of pyinstaller and setuptools are installed
-pip install -U "PyInstaller==3.2.1"
+pip install -U "PyInstaller"
```
singergr-ctds: Can we keep the pin here with the newer version? Assuming it's a newer version.

requirements.txt
```diff
@@ -1,10 +1,11 @@
...
-lxml==4.2.1
+lxml==4.4.2
 ndg-httpsclient==0.5.0
 pyasn1==0.4.3
 pyOpenSSL==18.0.0
 PyYAML==3.13
 progressbar2==3.43.1
 intervaltree==3.0.2
 termcolor==1.1.0
+requests==2.22.0
```
anmaxvl, line 11: maybe bump others too? e.g. pyyaml looks pretty old

ProfOak: I only upgraded what I had to in order to build on Windows without having to compile from source. I can look into upgrading some more, at least adding version ranges.

ProfOak: We discussed this a bit offline and decided to prioritize this at a later date to unblock qa.

anmaxvl (approved):

bin/package
singergr-ctds: I'm still not sure about removing this pin, but I'm okay with merging either way.

singergr-ctds (approved):

PR #250, opened 2019-12-13
fix(annotations): py3

scintillating7:
-bytes for py3
-clean up py2 test code
-add test to cover annotations

### DIFF
--- a/bin/gdc-client
+++ b/bin/gdc-client
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 import argparse
 import logging
--- a/tests/mock_server.py
+++ b/tests/mock_server.py
@@ -112,6 +112,7 @@ def download(ids=''):
     ids = ids.split(',')
 
     args = request.json
+
     if args:
         ids = args.get('ids')
 
@@ -135,7 +136,7 @@ def download(ids=''):
 
     if is_tarfile or is_compress:
         # make tarfile
-        make_tarfile(ids, filename)
+        make_tarfile(ids, filename, write_mode=write_mode)
 
         # load tarfile into memory to be returned
         with open(filename, 'rb') as f:
--- a/tests/test_download_client.py
+++ b/tests/test_download_client.py
@@ -1,10 +1,11 @@
 import copy
+from multiprocessing import Process, cpu_count
 import os
 import os.path
 import pytest
-import time
-from multiprocessing import Process, cpu_count
 import tarfile
+import tempfile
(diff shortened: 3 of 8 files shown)


tests/test_download_client.py
```diff
@@ -143,12 +142,47 @@ def test_download_tarfile(self):
...
+
+        # get annotation id out of metadata
+        index_client = GDCIndexClient(base_url)
+        index_client._get_metadata(files_to_dl)
+
+        # set expected output path for download client
+        override_kwargs = copy.deepcopy(client_kwargs)
+        override_kwargs['directory'] = os.getcwd() + '/tests'
+        # where we expect annotations to be written
+        output_path = override_kwargs['directory'] + '/{}/annotations.txt'.format(
```
anmaxvl: you should probably use pytest's tempfil/tmpdir fixtures for this, instead of creating an empty directory for tests to use. tests can fail for various reasons and you might not get to the cleanup section at all. as an alternative you can also do something around python's `tempfile` library (python3 has a context manager support for a directory)

tests/test_download_client.py
```diff
@@ -1,12 +1,10 @@
 import copy
 import os
 import os.path
-import tarfile
-import sys
-
 import pytest
 import time
 from multiprocessing import Process, cpu_count
+import tarfile
```
anmaxvl, line 6: since you're touching this, maybe reorder imports as well?

anmaxvl (changes requested):

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist=py27,py3
+envlist=py35,py36
```
anmaxvl, line 2: you probably want to update `.travis.yaml` and maybe extend pyenv to include 3.6 and 3.7?

anmaxvl (approved):
lgtm

philtom-ctds (approved):

bin/gdc-client
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
```
ProfOak, line 1: :+1:

tests/mock_server.py
```diff
@@ -135,7 +136,7 @@ def download(ids=''):
 
     if is_tarfile or is_compress:
         # make tarfile
-        make_tarfile(ids, filename)
+        make_tarfile(ids, filename, write_mode=write_mode)
```
ProfOak, line 139: :clap:

ProfOak (approved):

PR #251, opened 2019-12-16
fix(logging): respect logging format for py3

scintillating7:
Logging in py3 uses the format set in `style` rather than `format`. See [source code](https://github.com/python/cpython/blob/d730719b094cb006711b1cd546927b863c173b31/Lib/logging/__init__.py#L533).

### DIFF
--- a/gdc_client/log/log.py
+++ b/gdc_client/log/log.py
@@ -11,7 +11,7 @@ class LogFormatter(logging.Formatter):
     info_format = '%(msg)s'
 
 
-    def __init__(self, fmt='%(asctime)s - %(levelname)s: %(msg)s',
+    def __init__(self, fmt='%(asctime)s - %(levelname)s: %(msg)s', style='%',
                  color_off=False):
         logging.Formatter.__init__(self, fmt)
         self.color_off = color_off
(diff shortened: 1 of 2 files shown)


scintillating7:
Note: I tested this locally with `gdc-client download [uuid]`.

ProfOak (approved):
lgtm :+1:

philtom-ctds (approved):

scintillating7:
For future reference, we used to set the `logging.Formatter._fmt` attribute explicitly to `'%msg'`. In Python3 the `logging.Formatter._style` attribute is `{'_fmt': '%(asctime)s - %(levelname)s: %(msg)s'}` by default. 

The stacktrace for logging a record is roughly 
- `logging.Formatter.format()`
- checks [`usesTime`](https://github.com/python/cpython/blob/d730719b094cb006711b1cd546927b863c173b31/Lib/logging/__init__.py#L620)
- checks the `self._style` underlying attribute [here](https://github.com/python/cpython/blob/d730719b094cb006711b1cd546927b863c173b31/Lib/logging/__init__.py#L588)
- so since the `_style` contained time by default, we end up formatting it with the time [here](https://github.com/python/cpython/blob/d730719b094cb006711b1cd546927b863c173b31/Lib/logging/__init__.py#L433) and ignoring the simple `'%msg'` format we desired.

PR #253, opened 2019-12-23
feat(DTT-154): Remove pyinstaller workaround

ProfOak:
Prior to the recent Python3 upgrade we used an older version of the
requests library. This library came with a `cacert.pem` file needed to
make https connections. This file was separated out into it's own
library called `certifi`, and PyInstaller added a hook to include this
file on build time. The original workaround assumed the cacert.pem file
existed in the requests folder of the temp directory made by PyInstaller.

The workaround is no longer needed because of the PyInstaller hook.

Other changes in this commit revolve around Python3, and progressbar2 fixes.

- `e.message` is no longer needed in Python3.
- progress bar library was changed from progressbar to progressbar2 and
the behavior changed slightly.
- Reorganized imports in upload/client.py to pep8 recommendations.
- Added default parameter of True on verify request for download.

### DIFF
--- a/gdc_client/parcel/utils.py
+++ b/gdc_client/parcel/utils.py
@@ -47,11 +47,16 @@ def get_pbar(file_id, maxval, start_val=0):
 
     """
     log.debug('Downloading {0}:'.format(file_id))
-    pbar = ProgressBar(widgets=[
-        Percentage(), ' ',
-        Bar(marker='#', left='[', right=']'), ' ',
-        ETA(), ' ', FileTransferSpeed(), ' '], maxval=maxval, fd=sys.stdout)
-    pbar.currval = start_val
+    pbar = ProgressBar(
+        widgets=[
+            Percentage(), ' ',
+            Bar(marker='#', left='[', right=']'), ' ',
+            ETA(), ' ',
+            FileTransferSpeed(), ' ',
+        ],
+        initial_value=start_val,
+        maxval=maxval,
+        fd=sys.stdout)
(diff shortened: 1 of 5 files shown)


gdc_client/parcel/utils.py
```diff
@@ -51,7 +51,7 @@ def get_pbar(file_id, maxval, start_val=0):
         Percentage(), ' ',
         Bar(marker='#', left='[', right=']'), ' ',
         ETA(), ' ', FileTransferSpeed(), ' '], maxval=maxval, fd=sys.stdout)
-    pbar.currval = start_val
+    pbar.value = start_val
```
kulgan: Passing this in the constructor as `initial_value` might be more intuitive

ProfOak: Thanks for pointing this out. There are apparently numerous parameters for this class constructor that do not show up in the documentation or examples on their website. I had to go into the source code to figure it out.

https://github.com/WoLpH/python-progressbar/blob/develop/progressbar/bar.py#L276-L280

kulgan (approved):
lgtm. @ProfOak do you mind updating the branch, seems out of date with the base

anmaxvl (approved):
can you squash commits before we merge?

ProfOak:
> can you squash commits before we merge?

Correct me if I'm wrong but isn't there an option to do this on a merge? I'm not opposed to this. Please feel free to do this whoever has merge access. Let's get this merged soon, too. :-)

PR #257, opened 2020-04-20
DEV-71 add pre-commit and black to format code

asurinsaka:
Add Black and pre-commit in dev-requirements. 
    Format code by black. 
    Add instruction to install pre-commit in README.md.
    After installation, pre-commit will call black to format code 
    before each git commit autommatically.


ProfOak (approved):
That is a lot of double quotes changes. Thank you for taking care of this!

anmaxvl (approved):

PR #259, opened 2020-05-18
Feat/dev 161 remove duplicated code

asurinsaka:
refactor and remove duplicated code in tests/.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,21 @@
+from collections import namedtuple
 import hashlib
 from io import BytesIO
-import os
 import pytest
 import tarfile
 
 from gdc_client.parcel.const import HTTP_CHUNK_SIZE
+
+
+def generate_metadata_dict(access, contents, annotations, related_files):
+    return {
+        "access": access,
+        "contents": contents,
+        "file_size": None if contents is None else len(contents),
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -37,76 +50,44 @@ def make_tarfile(ids, tarfile_name="temp.tar", write_mode="w"):
...
+    ),
+    "small_ann": generate_metadata_dict(
+        "open", "small content 2", ["annotations.txt"], [],
+    ),
+    "small_rel": generate_metadata_dict("open", "small content 3", [], ["related 3"],),
+    "small_no_friends": generate_metadata_dict(
+        "controlled", "small content 4", [], [],
+    ),
+    "big": generate_metadata_dict(
+        "controlled", next(big_content), ["annotation 1"], ["related 1"],
```
anmaxvl: I think it'd make sense to make `big_content` a callable, instead of `generator`.

anmaxvl (approved):
just a minor comment. other than that looks good. thanks for cleanup

tests/conftest.py
asurinsaka: Thanks, I think you are right, use a callable makes much more sense here. Changed.

PR #260, opened 2020-05-18
DEV-139 give proper error message to user when uuid not found

asurinsaka:
remove debug message when the not in debug mode.


ProfOak (commented):
I tested this manually and can confirm that the stacktrace is gone. Can you extend the unit tests to include this case?

asurinsaka:
Thank you! Test added.

kulgan (approved):
lgtm

PR #261, opened 2020-05-28
Bug/dev 187 remove six

asurinsaka:
We do not need to support python 2 for gdc-client any more.

### DIFF
--- a/gdc_client/parcel/segment.py
+++ b/gdc_client/parcel/segment.py
@@ -30,7 +30,7 @@ from gdc_client.parcel.const import SAVE_INTERVAL
 
 if OS_WINDOWS:
     WINDOWS = True
-    from six.moves.queue import Queue
+    from queue import Queue
 else:
     # if we are running on a posix system, then we will be
     # communicating across processes, and will need
--- a/gdc_client/parcel/segment.py
+++ b/gdc_client/parcel/segment.py
@@ -15,8 +15,6 @@ import tempfile
 import time
 import sys
 
-from multiprocessing import Manager
-
 from intervaltree import Interval, IntervalTree
 from progressbar import ProgressBar, Percentage, Bar, ETA
 
@@ -29,6 +27,17 @@ from gdc_client.parcel.utils import (
     check_file_existence_and_size,
 )
 from gdc_client.parcel.const import SAVE_INTERVAL
+
+if OS_WINDOWS:
+    WINDOWS = True
+    from six.moves.queue import Queue
+else:
+    # if we are running on a posix system, then we will be
+    # communicating across processes, and will need
+    # multiprocessing manager
+    from multiprocessing import Manager
+
+    WINDOWS = False
 
 log = logging.getLogger("segment")
 
@@ -66,10 +75,13 @@ class SegmentProducer(object):
         self.block_size = work_size // self.n_procs
 
     def _setup_queues(self):
-
-        manager = Manager()
-        self.q_work = manager.Queue()
-        self.q_complete = manager.Queue()
+        if WINDOWS:
+            self.q_work = Queue()
+            self.q_complete = Queue()
+        else:
+            manager = Manager()
+            self.q_work = manager.Queue()
+            self.q_complete = manager.Queue()
 
     def integrate(self, itree):
         return sum([i.end - i.begin for i in itree.items()])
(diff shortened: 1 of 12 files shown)


gdc_client/parcel/segment.py
```diff
@@ -75,13 +66,10 @@ def _setup_work(self):
...
-            self.q_work = Queue()
-            self.q_complete = Queue()
-        else:
-            manager = Manager()
-            self.q_work = manager.Queue()
-            self.q_complete = manager.Queue()
+
+        manager = Manager()
+        self.q_work = manager.Queue()
+        self.q_complete = manager.Queue()
```
kulgan: I think the previous implementation used `Manager` only on posix systems due to windows lack of support for `fork()`. Not sure if this breaks anything but we should at least try it out on a windows system (that is if you have not already) before merging

asurinsaka: I download the source code and run in windows 10 with python 3.6. All the tests passed. 

To get the test passed, I have to modify the conftest.py file to use threading instead of Process from multiprocessing.https://github.com/NCI-GDC/gdc-client/blob/476103643d62aec943bfd0d6f05124f629aa8f4a/tests/conftest.py#L86-L95  Because multiprocessing. Process would try to pickle the flask app.run, which will fail. The Thread would never stop. Because there is no easy way to terminate it. 

There are 2 ways to solve this problem:
1. Some said on Stackoverflow that py3.7 will solve this
2. Pathos package use dill instead of pickle, which should work. 

In summary, if the test coverage is fine, the code should work for windows. The original tests from the develop branch will not work for windows anyway.

asurinsaka: I double-checked the test coverage. This part is not covered by any test. 
I run in debug mode manually and try to download a 100MB+ file and the modified version works for python 3.6 on Windows 10.

gdc_client/parcel/segment.py
```diff
@@ -28,17 +30,6 @@
 )
 from gdc_client.parcel.const import SAVE_INTERVAL
 
-if OS_WINDOWS:
-    WINDOWS = True
-    from six.moves.queue import Queue
```
philtom-ctds, line 33: To be safe, let's limit the changes in the PR to only the removal of `six`.    Let's just remove the `six` queue for the py3 `queue.Queue`.  Let's leave the windows specific code in place.

philtom-ctds (changes requested):

gdc_client/parcel/segment.py
```diff
@@ -30,6 +28,17 @@
 )
 from gdc_client.parcel.const import SAVE_INTERVAL
 
+if OS_WINDOWS:
+    WINDOWS = True
+    from six.moves.queue import Queue
```
philtom-ctds: Still need to remove the `six` reference here

scintillating7 (commented):
looks like you removed six (other than the queue @philtom-ctds mentioned)

gdc_client/parcel/segment.py
asurinsaka: From my understanding, the only reason the `WINDOWS` variable is used in this file is to decide whether to use the Queue from six.moves.queue or use the Queue from multiprocessing.Manager (line 78 of the file). It is not used anywhere else in this file.

I do not see a point to keep the windows code here and only remove the line of the six code. 

I think we should remove them or keep them all together. Could you please double-check this?

asurinsaka: And also, I can not understand why to use another `WINDOWS` variable when `OS_WINDOWS` is already used. It feels really wired for me to add the `WINDOWS` variable here.

If you think it makes sense to you, I will remove the six and keep the rest. I might be too new here to see the whole picture.

philtom-ctds: The purpose of the ticket is to remove the usage of `six`.  Limit your changes to that.  There are many places in this code base where we could make improvements.  Those are out-of-scope of this ticket.

I am not suggesting to delete this line of code entirely.  Instead, you should remove the use of `six` here and just use py3:
```
from queue import Queue
```

By doing this, we make a very straight forward change that should be 100% compatible with the existing functionality.

asurinsaka: Thanks, it makes sense.

philtom-ctds (approved):

scintillating7 (approved):
lgtm :)

kulgan (approved):

PR #262, opened 2020-06-12
DTT-159 DEV-228 multipart upload and replacing progressbar

anmaxvl:


gdc_client/download/parser.py
```diff
@@ -249,6 +249,7 @@ def config(parser, download_defaults):
         "frequent printout but lower performance.",
     )
     parser.add_argument(
+        "-k",
```
ProfOak, line 251: :clap:

gdc_client/download/parser.py
```diff
@@ -66,7 +66,7 @@ def download(parser, args):
 
     # Query the api to get the latest version of a file(s) according to the gdc.
     # Return OLD_ID => NEW_ID mapping
-    ids_map = get_latest_versions(args.server, ids)
+    ids_map = get_latest_versions(args.server, ids, verify=not args.no_verify)
```
ProfOak, line 68: :+1:

gdc_client/parcel/utils.py
```diff
@@ -39,30 +37,26 @@ def check_transfer_size(actual, expected):
...
-            " ",
-            ETA(),
-            " ",
-            FileTransferSpeed(),
-            " ",
-        ],
-        initial_value=start_val,
-        maxval=maxval,
-        fd=sys.stdout,
+    pbar = tqdm(
```
ProfOak: I don't remember but does the gdc-client use `#` to denote a progress bar? If so we can simulate that by using `ascii=True` as a parameter to tqdm. I anticipate this might come up during testing or user feedback.

gdc_client/parcel/segment.py
```diff
@@ -16,7 +16,7 @@
 import sys
 
 from intervaltree import Interval, IntervalTree
-from progressbar import ProgressBar, Percentage, Bar, ETA
+from tqdm import tqdm
```
ProfOak: Nice find. Do you think we should consider using this library for our other repositories like esbuild? I found it being used in data-tools a little, so it seems like something we might want in other places.

gdc_client/upload/client.py
```diff
@@ -78,47 +73,57 @@ def upload_multipart(
...
+                    chunk_file = mmap(
+                        fileno=source.fileno(),
+                        length=num_bytes,
+                        offset=offset,
+                        prot=PROT_READ,
+                    )
+
+                log.debug("Making http request for part {}".format(part_number))
+                res = requests.put(
+                    url + "?uploadId={0}&partNumber={1}".format(upload_id, part_number),
```
ProfOak: We can probably remove the numbers in these string formatters. That was a workaround to support python 2.6 on older CentOS systems, as implicit curly brace formatters were introduced in 2.7 (and py3k). I think the same CentOS systems shipped with python 3.3 as well, which supports implicit curly brace in formatted strings. Of course it doesn't hurt to keep them in.

> Changed in version 3.1: The positional argument specifiers can be omitted, so '{} {}' is equivalent to '{0} {1}'.

https://docs.python.org/3.3/library/string.html#formatstrings

gdc_client/upload/client.py
```diff
@@ -344,11 +347,6 @@ def load_file(self, file_entity):
         # Load attributes from a UploadFile to self for easy access
         self.__dict__.update(file_entity.__dict__)
 
-    # TODO: Is this used anywhere?
```
ProfOak, line 346: :heavy_check_mark:

ProfOak (commented):
Changes look good so far. Left a few minor questions.

gdc_client/parcel/segment.py
anmaxvl: yeah, I think it'd be a good replacement

gdc_client/parcel/utils.py
anmaxvl: good point

gdc_client/parcel/utils.py
```diff
@@ -156,9 +151,11 @@ def md5sum(block):
 
 def md5sum_whole_file(fname):
     hash_md5 = hashlib.md5()
-    with open(fname, "rb") as f:
+    with open(fname, "rb") as f, get_pbar(fname, None, desc="Validating md5sum") as pbar:
```
kulgan: Since its local file we can try to get the actual file size and use as max value for the progressbar using os.stats or something

tests/test_upload_client.py
```diff
@@ -1,20 +1,395 @@
...
+# This is based entirely on the known GraphQL query that is sent to backend
+GRAPHQL_RE = (
+    "^query \w+ \{ "
+    "(?P<type>\w+)"
+    " \(id: \""
+    "(?P<node_id>[a-z0-9-]+)"
+    "\"\) \{ "
+    "(?P<fields>[\w ]+)"
+    " } }$"
+)
```
philtom-ctds: Not everyone is going to be able to parse this regex, which will make it difficult for people to understand and maintain.  Let's hide this behind a function that describes the desired behavior.  That will be friendlier to a wider audience of developers, especially ones who are less experienced with regex.    For example,
```
QueryParts = namedtuple("QueryParts", ["type", "node_id", "fields"])
def parse_query(query: str) -> Optional[QueryParts]:
    # regex match
    # if no match, return None
    # return the query parts
```



https://xkcd.com/1171/

tests/test_upload_client.py
```diff
@@ -1,20 +1,395 @@
...
+@pytest.fixture
+def mock_files(mock_dir):
+    """Create 2 files: 10MB and 15MB in size"""
+    file_name1 = "fake_bam_1.bam"
+    file_name2 = "fake_bam_2.bam"
+
+    # Create 10MB file
+    f1 = mock_dir.join(file_name1)
+    size1 = 10 * 1024 * 1024
+    f1.write("a" * size1)
```
philtom-ctds, line 68: https://www.youtube.com/watch?v=ZWwMre0PuTQ

tests/test_upload_client.py
```diff
@@ -1,20 +1,395 @@
...
-        results = [
-            "/path/to/resume_file.yml",
-            "path/to/resume_file.yml",
-            "resume_file.yml",
-        ]
-        for i, t in enumerate(tests):
-            assert create_resume_path(t) == results[i]
+# This is based entirely on the known GraphQL query that is sent to backend
+GRAPHQL_RE = (
+    "^query \w+ \{ "
```
ProfOak: Should we use raw strings here to be safe? If it's not an escape character I think python will correctly encode it as `\\"`. If we use raw strings we remove all risk. I don't think it's an issue in these regex strings though.

```python
>>> "\w" == r"\w"
True
>>> "\n" == r"\n"
False

```

gdc_client/upload/client.py
```diff
@@ -510,73 +527,78 @@ def upload_parts(self):
...
+                        self.debug,
+                    ]
+                )
+            else:
+                self.total_parts -= 1
+
+        if self.total_parts == 0:
+            return
+
+        pbar = tqdm(
```
scintillating7: might be a good place to use `with` so you don't have to manually close the pbar

https://pypi.org/project/tqdm/#examples-and-advanced-usage

gdc_client/upload/client.py
```diff
@@ -65,7 +61,7 @@ def __getattr__(self, attr):
         return getattr(self._file, attr)
 
     def read(self, num):
-        self.pbar.update(min(self.pbar.value + num, self.filesize))
+        self.pbar.update(num)
```
scintillating7: is the min filesize logic no longer applicable?

gdc_client/download/client.py
```diff
@@ -281,22 +280,6 @@ def download_small_groups(self, smalls):
...
-                    ETA(),
-                    " ",
-                    FileTransferSpeed(),
-                    " ",
-                ],
-                maxval=1,
-                fd=sys.stdout,
-            )
-            pbar.start()
-
```
scintillating7, line 299: so is this download case for pbar completely covered by the one in gdc_client/parcel/segment.py?

scintillating7 (commented):
thank you for tests 🙏

ProfOak (approved):
The changes seem good. Thank you for adding more to the tests! I left one comment but I don't think it should block the merge.

tests/test_upload_client.py
```diff
@@ -1,20 +1,395 @@
...
+            "file_name": mock_files[0],
+        },
+        file_id2: {
+            "type": "submitted_unaligned_reads",
+            "project_id": "GDC-MISC",
+            "file_name": mock_files[1],
+        },
+    }
+
+    @httmock.urlmatch(netloc="localhost", method="POST", path="/v0/submission/graphql")
```
philtom-ctds, line 96: You weren't kidding when you said there was a lot of mocking to do.  Thanks for adding these to allow us to increase our test coverage.

For the long-term, we should think about how we want to perform integration tests for any of our code that depends on external services.  Mocking services like this fragile and problematic to maintain in the long-term.  
(not something we need to figure out for this pr though)

gdc_client/upload/client.py
anmaxvl: we're not calculating the actual value of the progress bar as we did before, but I guess updating with an actual size that's been read would be better here

gdc_client/download/client.py
anmaxvl: this progress bar was only to show the progress of grouping small files, I'm gonna push update to display the progress bar for the actual download

tests/test_upload_client.py
anmaxvl: with jenkins we should just make a pipeline for `gdc-client` specifically and talk to the deployed services. we could potentially mock more aggressively here, e.g. don't even have a `moto` or `httmock` and just mock `requests`, but meh

gdc_client/download/client.py
anmaxvl: done

tests/test_upload_client.py
philtom-ctds: I like the Jenkins idea.  I'll add it to my list of potential solutions for this testing problem.

philtom-ctds (approved):

gdc_client/upload/client.py
```diff
@@ -51,12 +51,8 @@ def __init__(self):
 log = logging.getLogger("upload-client")
 
 
-def upload_multipart_wrapper(args):
-    return upload_multipart(*args)
-
-
 class Stream(object):
-    def __init__(self, file, pbar, filesize):
+    def __init__(self, file, pbar: tqdm, filesize: int):
```
kulgan, line 54: I think this class is one of the problems with upload progress bar. `Request` demands a file-like object for streaming upload, see [docs](https://2.python-requests.org/en/master/user/advanced/#streaming-uploads) this class simply implements `read` with a parameter. I tried to verify that this is all `requests` requires to figure out if an object is file-like or not and it led me to this [implementation](https://github.com/psf/requests/blob/1ca1c52e698b13d3d5cc0755a6450306d880b933/requests/models.py#L471-L494)  which looks like it expects an `__iter__` and `tell` method instead of a `read` with a parameter. If this is correct, the way the class is implemented is why upload jumps from zero to completion.

gdc_client/download/client.py
```diff
@@ -276,27 +275,14 @@ def download_small_groups(self, smalls):
         errors = []
         groupings_len = len(smalls)
 
-        for i, s in enumerate(smalls):
+        for i, s in tqdm(iterable=enumerate(smalls), total=len(smalls),
```
kulgan: The progress bar above looks like it will still go from zero to complete for the small groups, since all the download event happen in this function. It might make sense to target the progress bar in the actual download function (self._download_tarfile(s)) instead so it captures the actual progress of the download

kulgan (changes requested):

kulgan (changes requested):
The change request is mainly for the `Stream` class implementation. Also, I thought we had coverage report for this repo on codacy

gdc_client/upload/client.py
anmaxvl: I don't think we use this class for multi-part uploads. not sure if the progress bar was an issue for regular non-multi-part uploads. I'll take a look

anmaxvl: turns out the `__getattr__` takes care of everything. the only concern I had is an implicit `__iter__` definition, but it doesn't look like `requests` is using it that way. I did manual pdb'ing to make sure. so we're good with the current implementation

gdc_client/parcel/segment.py
```diff
@@ -95,7 +95,7 @@ def validate_segment_md5sums(self, path=None):
         log.debug("Checksumming {0}:".format(self.download.url))
 
         pbar = tqdm(
-            intervals, desc="Segment md5sum validation", file=sys.stdout, ascii=True,
+            iterable=intervals, desc="Segment md5sum validation", file=sys.stdout,
```
ProfOak, line 98: This `tdqm` isn't a partial function. Do we still want to remove `ascii=True`?

anmaxvl: https://github.com/tqdm/tqdm#custom-integration and I did change the import https://github.com/NCI-GDC/gdc-client/pull/262/files#diff-8d934cd5147a38b2e9a623cdc45d3259R27

ProfOak: Oh I see. I missed that import statement. Carry on. :eyes:

kulgan (approved):

PR #264, opened 2020-06-22
feat(binary): Add github build binaries

kulgan:
Adds github action to build binaries on different os and some minor update to the packaging script

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - develop
+      - release/**
+      - hotfix/**
       - feat/dev-245-build-binaries
     tags:
       - "*"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,5 +44,5 @@ jobs:
       shell: bash
     - uses: actions/upload-artifact@v2
       with:
-        name: py${{ matrix.python }}-${{ env.GDC_CLIENT_ZIP }}
+        name: ${{ env.GDC_CLIENT_ZIP }}-py${{ matrix.python }}.zip
         path: bin/gdc-client_*.zip
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,6 @@ name: ci
 on:
   push:
     branches:
-      - master
       - develop
       - feat/dev-245-build-binaries
     tags:
@@ -18,8 +17,6 @@ jobs:
           - ubuntu-latest
           - windows-latest
         python:
-          - 3.5
-          - 3.6
           - 3.7
     env:
       GDC_CLIENT_ZIP: "gdc-client.zip"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,48 @@
+name: ci
+on:
+  push:
+    branches:
+      - master
+      - develop
+      - feat/dev-245-build-binaries
+    tags:
+      - "*"
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+        python:
+          - 3.5
+          - 3.6
+          - 3.7
+    env:
+      GDC_CLIENT_ZIP: "gdc-client.zip"
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Run Tests
+      if: matrix.os != 'windows-latest'
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox
+        tox -e py
+    - name: Package
+      run: |
+        cd bin
+        pip install virtualenv
+        . ./package
+        echo ::set-env name=GDC_CLIENT_ZIP::$GDC_CLIENT_ZIP
+      shell: bash
+    - uses: actions/upload-artifact@v2
+      with:
+        name: py${{ matrix.python }}-${{ env.GDC_CLIENT_ZIP }}
+        path: bin/gdc-client_*.zip
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,7 @@ jobs:
         os:
           - macos-latest
           - ubuntu-latest
+          - ubuntu-16.04
           - windows-latest
         python:
           - 3.7
@@ -43,5 +44,5 @@ jobs:
       shell: bash
     - uses: actions/upload-artifact@v2
       with:
-        name: ${{ env.GDC_CLIENT_ZIP }}-py${{ matrix.python }}.zip
+        name: ${{ env.GDC_CLIENT_ZIP }}-py${{ matrix.python }}-${{ matrix.os }}
         path: bin/gdc-client_*.zip
--- a/bin/package
+++ b/bin/package
@@ -105,15 +105,16 @@ echo "Testing produced binary..."
 echo "Zipping binary..."
 cd dist
 
-ZIP_NAME="gdc-client_${VERSION}_${TARGET_ENVIRONMENT}_x64"
+PRE_ZIP_NAME="gdc-client_${VERSION}_${TARGET_ENVIRONMENT}_x64"
 if [ -n "${DATE}" ]; then
-    ZIP_NAME="gdc-client_${VERSION}_${DATE}_${TARGET_ENVIRONMENT}_x64"
+    PRE_ZIP_NAME="gdc-client_${VERSION}_${DATE}_${TARGET_ENVIRONMENT}_x64"
 fi
 
-echo "The output zip file will be called ${ZIP_NAME}.zip"
-compress "$ZIP_NAME.zip" "$BINARY_NAME"
-mv "$ZIP_NAME.zip" ..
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+        python:
+          - 3.5
+          - 3.6
+          - 3.7
```
kulgan: Probably just pick one version instead

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+    env:
+      GDC_CLIENT_ZIP: "gdc-client.zip"
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Run Tests
+      if: matrix.os != 'windows-latest'
```
kulgan, line 33: Tests fail on windows, after a very quick assessment it looks like fixtures use multiprocessing which is not windows friendly. Would have loved to see tests pass on windows but that's for another day

.github/workflows/ci.yaml
ProfOak: I agree. 3.7 is probably a safe choice. There was some issue with wheels not being made for some of our pinned dependencies for py3.8 on Windows. It might go away if we upgrade the dependencies, assuming the newer deps have wheels built for them.

.github/workflows/ci.yaml
ProfOak: Is your vision for this that we migrate entirely from Travis to GitHub Actions? I wouldn't be opposed to that but I'd still bring it up with the rest of the team.

bin/package
```diff
@@ -102,14 +105,15 @@ echo "Testing produced binary..."
 echo "Zipping binary..."
 cd dist
 
-ZIP_NAME="gdc-client_${VERSION}_${TARGET_ENVIRONMENT}_x64.zip"
+ZIP_NAME="gdc-client_${VERSION}_${TARGET_ENVIRONMENT}_x64"
```
ProfOak: I'm not sure I understand this change. Can we keep the `.zip` extension in the `ZIP_NAME` variable and make a variable before this without the extension? Then the pre-name variable can be used in the rest of the action? 
```suggestion
PRE_ZIP_NAME="gdc-client_${VERSION}_${TARGET_ENVIRONMENT}_x64"
ZIP_NAME="$PRE_ZIP_NAME.zip"
```

We could also use bash built-in functionality to remove the `.zip` when we need it at the end.

```bash
$ ZIP_NAME="gdc-client-windows.zip"
$ echo ${ZIP_NAME%.zip}
gdc-client-windows
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
+name: ci
+on:
+  push:
+    branches:
+      - master
+      - develop
```
ProfOak, line 5: Can we limit this to the master branch only?

.github/workflows/ci.yaml
anmaxvl: is it upload by any chance? I added a threadpool executor 😆 maybe should've added processpool one

ProfOak (commented):
I had no idea we were working on this already but it's very cool! I left a few questions and minor suggestions.

.github/workflows/ci.yaml
anmaxvl: ~I think he wants to reduce human involvement in building the actual binaries, where github actions can take care of that for us~ ah sorry, I see what you mean, @ProfOak . My understanding is that there's no Windows pipeline to execute the tests at this moment. which we probably should... especially if something is broken now.

philtom-ctds: @ProfOak i imagine jenkins will be the travis replacement.

kulgan: @ProfOak I only wanted to build binaries and preserve the artifacts. Travis is still runs, I wasn't thinking about removing travis. @anmaxvl see https://github.com/NCI-GDC/gdc-client/runs/796762713?check_suite_focus=true

bin/package
kulgan: This change is mainly to power this command https://github.com/NCI-GDC/gdc-client/blob/9e1957b56493a3975721f625e1abefca59bef512/.github/workflows/ci.yaml#L43
which sets the artifact name. I agree, your implementation above is neater

.github/workflows/ci.yaml
kulgan: @ProfOak To be clearer, my goal was just to use actions build binaries for different os and preserve the artifacts here on github and possibly add them to the release. I added tests execution because it didn't increase the overall execution time by much, but yea I see your point as it makes travis somewhat redundant now. We can discuss that later if we want

.github/workflows/ci.yaml
philtom-ctds: This should probably be `release/**`, `hotfix/**`, and maybe `develop`.  This will allow us to build clients for a given release/hotfix branch.  Then, once the release/hotfix goes out, we can merge the branch down to `master`.  

Including `develop` would allow us to have testable artifacts after every merge, which sounds nice in theory, but could cause some bloat issues.  (see my comment about upload).  

`master` should be excluded here, as we don't want to overwrite the artifacts that actual got released.

kulgan: probably `develop` master is out of sync

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+      run: |
+        cd bin
+        pip install virtualenv
+        . ./package
+        echo ::set-env name=GDC_CLIENT_ZIP::$GDC_CLIENT_ZIP
+      shell: bash
+    - uses: actions/upload-artifact@v2
+      with:
+        name: ${{ env.GDC_CLIENT_ZIP }}-py${{ matrix.python }}.zip
+        path: bin/gdc-client_*.zip
```
philtom-ctds: Does this commit the build artifact into the repository?  If so, that can cause some bloat in our repo.  Git is not a good place to store binary data.  Are there actions that would allow us to upload the artifacts some where else?  Or even to a different repo?

Each build is ~45 MB in total.   We don't do that many dtt releases.  So, if we committed these for each release, we'd slowly grow the repos size.  16 releases over five years.  That's ~150 MB per year.  Over time, it'll add up though.

But if we committed artifacts for every merge to `develop`, then we'll grow much faster.  395 commits over 5 years.  That 3.5 GB per year.

.github/workflows/ci.yaml
ProfOak: Oh wow yeah it is.

philtom-ctds (commented):
This is pretty awesome!  Much better than manually building on three different machines.  My only concern here is bloating the repo with the binary artifacts.

.github/workflows/ci.yaml
ProfOak: @philtom-ctds that was my impression as well. I just wanted to make sure.

One of the "on" events for GitHub actions is when a release tag is created. That might be easier to manage since we don't normally build binaries until we're getting ready to release. It would let another tool (Travis or Jenkins) handle the test execution part, too.

https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onevent_nametypes

.github/workflows/ci.yaml
kulgan: The artifacts are not committed to the repo, they are uploaded to a public repository for public repos on github, See https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts for more details and expiry

ProfOak: We could do long term storage for artifacts on Nexus. I thought we might originally build on Jenkins and store on Nexus prior to this PR. I'm not entirely sure how that process would look like though.

philtom-ctds (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - "*"
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
```
singergr-ctds, line 19: I don't know how glibc compatibility works on Ubuntu, so I don't know if `ubuntu-latest` might cause any issues, but if the binary works on Ubuntu 16.04, we should be good.

philtom-ctds: I wondered that same thing.  I figured we can try this and test out the client.  If there's issues, then we can submit another pr to use `ubuntu-16.04` instead

anmaxvl (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+      if: matrix.os != 'windows-latest'
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox
+        tox -e py
+    - name: Package
+      run: |
+        cd bin
+        pip install virtualenv
+        . ./package
```
singergr-ctds, line 42: This seems a little too clever, but it works in this context, so. 🤷‍♂️

singergr-ctds (approved):
Yeah, this is gonna be a lot better.

.github/workflows/ci.yaml
kulgan: You what @singergr you really do detect the tiniest of details. Good eyes :+1: . I should really give credit to the stackoverflow answer that pointed me to exec it like this so the environment variable can be kept long enough to pass it to github action

PR #268, opened 2020-07-01
bring back progressbar for multipart upload and download

anmaxvl:


gdc_client/parcel/utils.py
```diff
@@ -51,12 +64,43 @@ def get_pbar(file_id, maxval, start_val=0, desc="Downloading"):
         desc: display message next to the file_id
```
philtom-ctds: Could you update this doc to state that it's used for debug logging now?

requirements.txt
```diff
@@ -9,3 +9,4 @@ intervaltree==3.0.2
 termcolor==1.1.0
 requests==2.22.0
 tqdm==4.46.1
```
philtom-ctds: We can remove `tqdm`.

gdc_client/download/client.py
```diff
@@ -274,9 +274,10 @@ def download_small_groups(self, smalls):
         errors = []
         groupings_len = len(smalls)
 
-        for i, small_group in tqdm(
-            iterable=enumerate(smalls), total=len(smalls), unit="group"
-        ):
+        pbar = get_percentage_pbar(len(smalls))
```
philtom-ctds: This one used to have the `FileTransferSpeed` too:
https://github.com/NCI-GDC/gdc-client/blob/7764d84e43a0f2387b4f438c913e8af58d32ba8b/gdc_client/download/client.py#L284-L297

Does it render the same or differently?

Edit:  Actually, it used to do a progress bar per group instead of a single progress bar for all groups.  Personally, I think this way makes more sense.   But given the resistance we've seen, let's revert to the old way.

gdc_client/parcel/segment.py
```diff
@@ -317,7 +317,7 @@ def wait_for_completion(self):
                     this_size = interval.end - interval.begin
                     self.size_complete += this_size
                     since_save += this_size
-                    self.pbar.update(this_size)
+                    self.pbar.update(self.size_complete)
```
philtom-ctds: The previous code has a `try-catch` here because `pbar.update(n)` will raise an error if `n > pbar.max_value`.  We should either add that `try-catch` back or have some logic to make sure we're updating with a valid value.  Having the progresss bar abort a transfer would be undesirable.

gdc_client/upload/client.py
```diff
@@ -575,7 +577,8 @@ def upload_parts(self):
             for future in as_completed(future_to_part_number):
                 part_number = future_to_part_number[future]
                 log.debug("Part: {} is done".format(part_number))
-                pbar.update()
+                pbar.update(self.ns.completed)
```
philtom-ctds, line 580: This is much nicer.  Single-threaded updates to the progress bar 👍

philtom-ctds (commented):
I compared this PR to the old code.  Left a few comments about differences.  Could you also update the commit message to reference ticket DTT-169?

requirements.txt
anmaxvl: I haven't pulled it out from md5sum validation. I can do that too I guess. I'll have to double check

gdc_client/download/client.py
anmaxvl: revert back to a progressbar with maxval of `1`? I don't think it makes sense. especially having a `FileTransferSpeed` there, because the maxval doesn't represent the file size

philtom-ctds: Yes, let's revert.  I agree that your approach is better than the original approach.  But at this point, I don't want to give anyone any reasons to continue pushing back on progress bar changes.  I'd much rather get the upload fixes released, as that's a real problem affecting users.

gdc_client/download/client.py
```diff
@@ -274,16 +274,16 @@ def download_small_groups(self, smalls):
...
         for i, small_group in enumerate(smalls):
-            pbar.update(i + 1)
 
             if not small_group:
                 log.error("There are no files to download")
                 return [], 0
 
             log.debug("Saving grouping {0}/{1}".format(i + 1, groupings_len))
+
+            pbar = get_percentage_pbar(1)
```
anmaxvl, line 285: I'm not using a file progress bar, because it'll show the speed as `1.7b/sec` (or some other nonsense) due to the `max_value=1`.

scintillating7: didn't we used to show that nonsense tho? 😆

gdc_client/parcel/segment.py
```diff
@@ -94,10 +94,11 @@ def validate_segment_md5sums(self, path=None):
 
         log.debug("Checksumming {0}:".format(self.download.url))
 
-        pbar = tqdm(iterable=intervals, file=sys.stdout)
+        pbar = get_percentage_pbar(len(intervals))
 
         with mmap_open(path or self.download.path) as data:
-            for interval in pbar:
+            for i, interval in enumerate(intervals):
+                pbar.update(i + 1)
```
scintillating7: I think this is a little different where we used to update with the size_complete? and it was later in the flow? https://github.com/NCI-GDC/gdc-client/pull/262/files#diff-8d934cd5147a38b2e9a623cdc45d3259L284 I could be reading this wrong though.

gdc_client/upload/client.py
```diff
@@ -431,7 +432,7 @@ def _upload(self):
                     log.error("Can't upload: {}".format(r.content))
                     return
 
-                pbar = tqdm_file(total=self.file_size)
+                pbar = get_file_transfer_pbar(self.file_path, self.file_size, desc="Uploading")
```
scintillating7, line 435: did the old behavior have the "Uploading" prefix?

philtom-ctds: That `desc` is only used in debug logging, not the progress bars themselves.   So, it's ok.  The additional debug logging is beneficial.

gdc_client/parcel/segment.py
anmaxvl: this is a different progressbar. it's for md5sum validation, the one you're pointing to is for file download. this shows the progress of how many segments have been validated

philtom-ctds (approved):

PR #269, opened 2020-09-21
DEV-401 Fix multipart upload progress logging

jiakf:
When part fails for multipart upload,
progressbar is no longer updated.
The additional retry of the whole multipart upload
is redundant and leads to misleading logging,
so it is removed.


jiakf:
The retry of the whole multipart upload seemed redundant, as each part is already retried 10 times before giving up. It also leads to misleading logging, so I removed it for now. 

For the tests, I added tests for simple uploads as well, as well as a test for when a part of the multipart upload fails. There might be a better way to reorganize the tests, I tried refactoring a bit and I also had a version with parameterization, but that blended in logic between simple and multipart uploads in mock server. Open to suggestions on how to better refactor the tests. 

Also, a thing I noticed is that Debug=True is passed into the upload client in the multipart tests. I believe the intended behavior for DTT upload is when one file fails, it will continue to the next one. With Debug=True for multipart, as soon as a file part fails to upload, it doesn't continue to the next file and terminates. In Production, Debug is always False, so I changed it to false in the tests to better mimic production.

kulgan (commented):
lgtm, still going through all the changes in the tests

tests/test_upload_client.py
```diff
@@ -296,6 +327,27 @@ def handle_put_parts(url, req):
...
+    client = boto3.client("s3")
+    key = get_key(url.path)
+
+    # for simple upload, return 200 for initial dry_run request
+    if "dry_run" in url.path:
+        return httmock.response(200, "OK")
+
+    # simple upload
+    data = req.body.read(req.body.filesize)
+    result = client.put_object(Body=data, Bucket="test-bucket", Key=key)
```
ProfOak: Did we want to use `result` anywhere or can we get rid of it? I don't think there's anything from `put_object` that directly indicates a successful upload so it's probably ok to assume it was successful like we're doing now.

ProfOak (approved):

gdc_client/upload/client.py
```diff
@@ -452,12 +459,10 @@ def multipart_upload(self):
                 self.upload_parts()
                 self.check_multipart()
 
-                # try again in case some parts failed
-                if self.ns.completed != self.total_parts:
-                    self.upload_parts()
```
philtom-ctds, line 466: Could you explain why this was removed?

gdc_client/upload/client.py
```diff
@@ -564,8 +570,11 @@ def upload_parts(self):
 
             for future in as_completed(future_to_part_number):
                 part_number = future_to_part_number[future]
-                log.debug("Part: {} is done".format(part_number))
-                pbar.update()
+                if future.result():
```
philtom-ctds, line 565: Add a comment to note two things:
1.  The result of this future should be a boolean indicating success or failure.  
2. The wrapped call catches all exceptions.  so, an exception should not be re-raised here.

tests/test_upload_client.py
```diff
@@ -38,9 +35,39 @@ def parse_graphql_query(query: str) -> Optional[QueryParts]:
...
+        files=[
+            {"id": "file-id-2", "path": mock_dir_path},
+            {"id": "file-id-1", "project_id": "GDC-MISC", "path": mock_dir_path},
+        ],
+        debug=False,
+    )
+
+
+@pytest.fixture
+def mock_mp_upload_client(mock_dir_path):
```
philtom-ctds: Could you expand "mp"?  This stands for "multipart", right?   Same for the tests with "mp" in their name.

I can guess that mp == multipart in the context of this PR.  But it'd be harder to figure that without the context.

gdc_client/upload/client.py
```diff
@@ -564,8 +570,11 @@ def upload_parts(self):
...
+                    log.debug("Part: {} is done".format(part_number))
+                    pbar.update()
+                else:
+                    log.warning("Part: {} failed".format(part_number))
```
philtom-ctds, line 570: Since the key change in this pr is this logging change, could you add a test that captures the logs and verifies that the correct message is logged?  You should be able to use [caplog](https://docs.pytest.org/en/stable/logging.html#caplog-fixture) to test this.

gdc_client/upload/client.py
jiakf: so there are two reasons why I removed it:

1. the retry is redundant
each part is already retried up to MAX_RETRIES (which is 10 at the moment), so retrying the whole operation again for the failed parts seems arbitrary/not very useful (if we wanted to, we could make it a configurable thing, where we could set it to retry X number of times instead of just once). 

2. The log is confusing with the second call to self.upload_parts 
with my changes, if we take a file broken into 72 parts, and one part fails on the first upload_parts() call,  complete() will throw the exception saying "Multipart upload failed for file {}: completed parts: {}, total parts: {}, please try to resume", which is caught in handle_multiparts() and logged. So this would say 71 parts are completed out of 72 completed parts. 

The upload_parts() function seems functional in the sense that it is only capturing how many parts are successful out of the total parts tried in this particular run, but the problem arises when we call upload_parts() a second time (the current implementation). The logic in upload_parts() basically selects the parts that fails and uses the number of failed parts as the new total parts. Since this second upload_parts() gets called again before complete() gets called, we'll see that with this example, we'll see in the log that 0 parts were completed out of 1 total part, which is misleading. The problem is we're trying to use upload_parts(), which isn't necessarily in charge of keeping track of the overall successful parts out of the completed parts, but just on a particular run, out of the parts we want to try, how many were successful.

tests/test_upload_client.py
jiakf: Yes, i was following the convention that was already there when they were using "mp" to signify multipart. I can expand them to multipart

kulgan (approved):

gdc_client/upload/client.py
```diff
@@ -1,23 +1,23 @@
...
 import random
+import requests
 import time
+import yaml
+
 from collections import deque
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from gdc_client.parcel.utils import tqdm, tqdm_file
+from gdc_client.upload import manifest
+from lxml import etree
```
scintillating7: 👍

gdc_client/upload/client.py
```diff
@@ -180,26 +180,24 @@ def __init__(
         self.upload_id = None
         self.debug = debug
         self.processes = processes
-        self.upload_part_size = int(
-            max(upload_part_size, MIN_PARTSIZE) / PAGESIZE + 1
-        ) * PAGESIZE
+        self.upload_part_size = (
+            int(max(upload_part_size, MIN_PARTSIZE) / PAGESIZE + 1) * PAGESIZE
+        )
```
scintillating7: did you make this change because it's possible for `PAGESIZE` to not be an int?

gdc_client/upload/client.py
```diff
@@ -410,7 +415,9 @@ def delete(self):
             if r.status_code == 204:
                 log.info("Delete file {}".format(self.node_id))
             else:
-                log.warning("Fail to delete file {}: {}".format(self.node_id, r.content))
+                log.warning(
+                    "Fail to delete file {}: {}".format(self.node_id, r.content)
+                )
```
scintillating7: Want to use f-strings because we're on python3 now?

tests/test_upload_client.py
```diff
@@ -1,23 +1,20 @@
-from collections import namedtuple
+import boto3
+import httmock
 import json
 import os
+import pytest
```
scintillating7: pytest is third-party 😄

tests/test_upload_client.py
```diff
@@ -38,9 +35,39 @@ def parse_graphql_query(query: str) -> Optional[QueryParts]:
...
+        processes=2,
+        server="localhost",
+        upload_part_size=FIVE_MB,
+        multipart=True,
+        files=[
+            {"id": "file-id-2", "path": mock_dir_path},
+            {"id": "file-id-1", "project_id": "GDC-MISC", "path": mock_dir_path},
+        ],
+        debug=False,
     )
```
scintillating7: might be nice to use a [partial](https://docs.python.org/3.7/library/functools.html#functools.partial) to freeze the parameters that aren't `multipart` to cut down on duplicate code.

tests/test_upload_client.py
```diff
@@ -296,18 +327,37 @@ def handle_put_parts(url, req):
...
+
+    # simple upload
+    data = req.body.read(req.body.filesize)
+    client.put_object(Body=data, Bucket="test-bucket", Key=key)
+    return httmock.response(200, "OK")
+
+
 @httmock.urlmatch(netloc="localhost", method="DELETE")
-def handle_delete_mp(url, _):
+def handle_delete_multipart(url, _):
```
scintillating7, line 345: +1 for spelling things out

scintillating7 (commented):
Nice work.

gdc_client/upload/client.py
jiakf: i think this is just a black formatting change

gdc_client/upload/client.py
jiakf: so this was the thing i was thinking of during the pre-commit meeting right now, basically i had f-strings here, and then when i pushed, it failed the build check because it is running against python 3.5 also
https://github.com/NCI-GDC/gdc-client/blob/develop/.travis.yml#L15
so I reverted to the old way of formatting

tests/test_upload_client.py
jiakf: i refactored to use a helper function that will return a constructed client based on a passed in boolean for whether it is a multipart or not, I didn't think using partial seemed to apply here, but if the helper function isn't as succinct as using a partial, i can change it.

jiakf:
The counter variable being used in Manager Namespace() member variable
was being passed into each process, and the update was not being
synchronized among the processes, causing the completed variable to be
incorrect sporadically. There doesn't seem to be any need for the
manager namepsace, as we can update the counter variable as each future returns. 

The Manager object is created here for non windows os:
https://github.com/NCI-GDC/gdc-client/blob/develop/gdc_client/upload/client.py#L528

The FakeNamespace for windows is created here that initializes the counter variable:
https://github.com/NCI-GDC/gdc-client/blob/develop/gdc_client/upload/client.py#L42

So in upload_multipart, which is the function that is passed into each thread, the namespace variable is passed into each thread, and if the upload was successful, it will increment the shared counter variable in the thread. 
https://github.com/NCI-GDC/gdc-client/blob/develop/gdc_client/upload/client.py#L117

In the complete check, the counter variable is checked to match if it equals the total parts that were in the multipart upload
https://github.com/NCI-GDC/gdc-client/blob/develop/gdc_client/upload/client.py#L585

This is where i was noticing my test failing. Basically, I was observing that even though all parts uploaded successfully, when the test failed, the self.ns.completed value wasn't correct. I'm not sure what the reasoning is as to why we needed the Namespace class at all, as we update the progress bar in the main thread when futures return one by one, so the update to the progress bar should be tied to this counter variable, too.

gdc_client/upload/client.py
```diff
@@ -564,8 +553,17 @@ def upload_parts(self):
...
-                log.debug("Part: {} is done".format(part_number))
-                pbar.update()
+                """
+                    upload_multipart() returns True on success or False on failure
+                    upload_multipart() is responsible for catching exceptions,
+                    so no exception should be re-raised here
+                """
+                if future.result():
+                    log.debug("Part: {} is done".format(part_number))
+                    self.completed += 1
```
kulgan, line 567: :100: I also prefer updating `completed` in this manner, its more intuitive and probably better in terms of thread safety.

gdc_client/upload/client.py
```diff
@@ -554,8 +542,9 @@ def upload_parts(self):
         if self.total_parts == 0:
```
kulgan: Can we do something better for the sake of reporting totals and completed when upload finishes. We can add another variable, say `previously_uploaded` and increment that instead of decrementing the `total_parts`, that way `total_parts` always remains the same and makes sensible reports in scenarios where the user doe a `resume` for a previously in-progress upload. This line becomes something like this
``` python
if self.total_parts == previously_uploaded:
    return
```
We will have to update this to reflect previously uploaded. https://github.com/NCI-GDC/gdc-client/blob/251dcb4ca104bfcf4d2ca4ea1b4dc761d8ff660f/gdc_client/upload/client.py#L584-L588

gdc_client/upload/client.py
```diff
@@ -38,11 +37,6 @@
 
     from multiprocessing.pool import ThreadPool as Pool
```
kulgan: I think this can be safely removed.

This might make sense for a different investigation ticket/task. We might be able to safely remove the `freeze_support()` call since we switched to `concurrent.futures.ThreadPoolExecutor` which does not use the `multiprocessing` I believe. The `concurrent.futures.ProcessPoolExecutor` uses the `multiprocessing` module so a call to `freeze_support()` on windows would still be needed.

gdc_client/upload/client.py
jiakf: Ok, so I added a variable called previously_uploaded, which keeps track of how many parts have been already uploaded in upload_parts. I then check if previously_uploaded == part_amount (if all parts have already been uploaded, just return). only after we get past that check do I set self.total_parts, which now represents the number of actual parts that are going to be attempted to be uploaded, if that makes sense.

PR #273, opened 2021-01-04
Feat/dev 505 json response error

jiakf:


gdc_client/query/versions.py
```diff
@@ -30,14 +30,23 @@ def get_latest_versions(url, uuids, verify=True):
...
+                    # Might happen for legacy files
+                    latest_versions[file_id] = file_id
+        except json.JSONDecodeError as e:
+            # json() will fail if response body is not valid JSON
+            logger.error(
+                "The following response content resulted from request {0} for ids {1}: \n{2}".format(
+                    versions_url, chunk, resp.content
+                )
+            )
+            raise e
```
BenX00: You should just use raise here to re-raise the exception. That way the original stack trace is preserved.

tests/test_query.py
```diff
@@ -115,3 +117,16 @@ def test_get_latest_versions(
...
+@pytest.mark.parametrize(
+    "ids, latest_ids", [(["foo", "bar"], ["foo", "baz"])],
+)
+def test_get_latest_versions_error(
+    versions_response_error, ids: List[str], latest_ids: List[str],
+) -> None:
+    url = "https://example.com"
+    versions_response_error(url + "/files/versions")
+
+    with pytest.raises(Exception):
```
BenX00: Shouldn't we make sure that it's throwing the expected JSONDecodeError? Also this might be a good time to use Mocks to insure that the logger.error was at least called too.

gdc_client/query/versions.py
```diff
@@ -30,14 +30,23 @@ def get_latest_versions(url, uuids, verify=True):
...
         # Parse the results of the chunked query.
-        for result in resp.json():
-            file_id = result.get("id")
-            uuid = result.get("latest_id")
-            if uuid:
-                latest_versions[file_id] = uuid
-            else:
-                # Might happen for legacy files
-                latest_versions[file_id] = file_id
+        try:
```
BenX00: Is there a reason we opted not to check the status of the response to insure that we have a 2XX response before trying to deserialize the body? From the test that's what it looks like we are trying to solve for. If that is our main concern we could handle this with an if/else.

gdc_client/query/versions.py
jiakf: the original stack trace is preserved either way: both "raise e" and "raise" give me the same output when testing. I think its only when you raise a new Exception that you lose the original stack trace

gdc_client/query/versions.py
jiakf: i didn't find a good way to check if the response content is valid json: from what I've seen online, its suggested to just do a try/catch to see if its valid json. I was thinking about using the Response class's ok flag, which will return true if the response code is less than 400. Also, the current behavior is this request fails, to basically stop execution of the download, and re-raising the error will propogate up to the top level try/catch and end execution immediately with this error message. I could do the same thing with an if check to check the status code and just raising an error, so I could also do that with the Response class's ok flag, although now there wouldn't be a notion of using JSONDecodeError, as we'll see just the response content and status code

gdc_client/query/versions.py
```diff
@@ -29,6 +29,17 @@ def get_latest_versions(url, uuids, verify=True):
     for chunk in _chunk_list(uuids):
         resp = requests.post(versions_url, json={"ids": chunk}, verify=verify)
 
+        # anything greater than or equal to a 400 status code is not allowed
+        if not resp.ok:
+            raise Exception(
```
BenX00: This reads much better. The only nit picky things I have are A) could we use a more precise error? Either built-in or custom. And B) is the above comment needed? 

I understand the fear around not getting back a valid json. However, if we got back a invalid json from the service for an ok response where we requested json, that would mean the service was completely broken and I don't see that as likely. Plus, we'd prob know because we'd be having failures in a lot more places than this. The 502 isn't a valid json because that level of response isn't necessarily coming from the service itself at all.

jiakf: it seems like the norm for dtt to just throw generic Exception for every error (for better or worse). I added the comment because I didn't know that resp.ok is True for both 2XX and 3xx errors (I would've just thought 2XX errors). Could remove it though if its not needed

BenX00: Ok is a pretty universal 400 or less because 300s tech are errors they are redirects. If you are worried about 300s tho, you can always do an explicit check of the status code to make sure its in the 200's instead of using the ok prop. 

As for the more applicable exception, this shouldn't cause any problem at all. It just makes the expected behavior of this code more explicit. Also you should make note of the exception that is raised in the doc string.

tests/test_query.py
```diff
@@ -115,3 +117,19 @@ def test_get_latest_versions(
...
 
     assert result == expected
+
+
+@pytest.mark.parametrize("ids", [(["foo", "bar"])])
+def test_get_latest_versions_error(versions_response_error, ids: List[str],) -> None:
+    url = "https://example.com"
+    versions_response_error(url + "/files/versions")
+
+    with pytest.raises(Exception) as e:
```
BenX00: Using a more nuanced exception also allows this part of the test to be a little more meaningful as it will be looking for an expected exception type. Another note you can use the match property to do the checking of the error message instead of doing it below.

gdc_client/exceptions.py
```diff
@@ -1,3 +1,6 @@
 class ClientError(Exception):
-    """ Base client error.
```
ProfOak, line 2: :+1:

gdc_client/query/versions.py
```diff
@@ -29,6 +30,16 @@ def get_latest_versions(url, uuids, verify=True):
     for chunk in _chunk_list(uuids):
         resp = requests.post(versions_url, json={"ids": chunk}, verify=verify)
 
+        if not resp.ok:
+            raise ServerError(
+                (
+                    "The following request {0} for ids {1} returned with "
+                    "status code: {2} and response content: {3}"
```
ProfOak, line 38: I don't think this needs to change but I'd like to provide some historical context for why you might see this syntax in the repo. Previously when this repo was a Python 2 code base we had to support 2.6 for some of our users who couldn't upgrade their PCs. Python 2.6 had support for `.format` but the args had to be named in some way. Either using `"{name}.format(name="Jerry")` or `"{0}".format("George")`. We've since upgraded to Python 3 so I don't believe this is necessary any more.

ProfOak (approved):

XukunXiangUChicago (approved):
lgtm

gdc_client/exceptions.py
BenX00, line 2: I def like the idea here! Just have to say that these should prob be exceptions vs errors. Unless you are saying these are exceptions that we could never recover from.

gdc_client/query/versions.py
```diff
@@ -29,6 +30,16 @@ def get_latest_versions(url, uuids, verify=True):
     for chunk in _chunk_list(uuids):
         resp = requests.post(versions_url, json={"ids": chunk}, verify=verify)
 
+        if not resp.ok:
+            raise ServerError(
```
BenX00: One thing is now that you've split this into Client/Server exceptions/errors this isn't always the correct logic. The 400s are client errors and only the 500s are server errors. But because ok is >=400 you are throwing the same for each.

BenX00 (commented):
Have a minor comment

gdc_client/exceptions.py
jiakf: ClientError already existed, I just added the ServerError one.  The code right now suggests that if either of these errors occur, it'll basically end the program. If we wanted to change the name of ClientError, should be in another ticket, as it would touch other parts of the codebase.

gdc_client/query/versions.py
jiakf: I was looking for a generic name for an exception that resulted from this failed http request in general. I can change it to something like HTTPError or something similar.

gdc_client/exceptions.py
```diff
@@ -1,3 +1,6 @@
...
-    """
+    """Base client error."""
+
+
+class HTTPError(Exception):
+    """Base HTTP error."""
```
BenX00: We should prob capture the status code of the response for this as a property of this exception. And I would suggest making this an exception vs an error.

BenX00 (approved):
Looks good other than one last minor thing.

kulgan (approved):

PR #276, opened 2021-01-12
Feat/dev 506 large file download

jiakf:
details can be found in the ticket as to what problems are addressed here:
https://jira.opensciencedatacloud.org/browse/DEV-506

One thing that would change with this PR is the state of the partial file. Before this change, even if a download failed, the partial file would be renamed to the final file name. With this change, in the event that all the download retries of a particular file failed, the partial file would be left in the directory and not renamed. The next run of DTT would see this file, see that it is incomplete, and load state accordingly. The full scope of the change can be found in the ticket

### DIFF
--- a/gdc_client/common/config.py
+++ b/gdc_client/common/config.py
@@ -40,7 +40,7 @@ class GDCClientConfigShared(object):
         "n_processes": ConfigParser.getint,
         "retry_amount": ConfigParser.getint,
         "wait_time": ConfigParser.getfloat,
-        "no_segment_md5sum": ConfigParser.getboolean,
+        "no_segment_md5sums": ConfigParser.getboolean,
         "no_file_md5sum": ConfigParser.getboolean,
         "no_verify": ConfigParser.getboolean,
         "no_related_files": ConfigParser.getboolean,
@@ -49,6 +49,7 @@ class GDCClientConfigShared(object):
         "insecure": ConfigParser.getboolean,
         "disable_multipart": ConfigParser.getboolean,
         "path": ConfigParser.get,
+        "latest": ConfigParser.getboolean,
     }
 
     def __init__(self, config_path=None):
@@ -71,7 +72,7 @@ class GDCClientConfigShared(object):
                 "dir": ".",
                 "save_interval": SAVE_INTERVAL,
                 "http_chunk_size": HTTP_CHUNK_SIZE,
-                "no_segment_md5sum": False,
+                "no_segment_md5sums": False,
                 "no_file_md5sum": False,
                 "no_verify": False,
                 "no_related_files": False,
@@ -79,6 +80,7 @@ class GDCClientConfigShared(object):
                 "no_auto_retry": False,
                 "retry_amount": 1,
                 "wait_time": 5.0,
+                "latest": False,
             },
             "upload": {
                 "path": ".",
--- a/gdc_client/parcel/client.py
+++ b/gdc_client/parcel/client.py
@@ -110,7 +110,7 @@ class Client(object):
 
         log.debug("Download complete" + rate_info)
 
-    def validate_file_md5sum(self, stream):
+    def validate_file_md5sum(self, stream, file_path):
 
         if not stream.check_file_md5sum:
             log.debug("checksum validation disabled")
@@ -126,7 +126,7 @@ class Client(object):
                 "Cannot validate this file since the server did not provide an md5sum. Use the '--no-file-md5sum' option to ignore this error."
             )
 
-        if utils.md5sum_whole_file(stream.path) != stream.md5sum:
+        if utils.md5sum_whole_file(file_path) != stream.md5sum:
             raise Exception("File checksum is invalid")
 
     def download_files(self, urls, *args, **kwargs):
@@ -158,10 +158,11 @@ class Client(object):
 
             # Download file
             try:
+                # validate temporary file before renaming to permanent file location
                 self.parallel_download(stream)
+                self.validate_file_md5sum(stream, stream.temp_path)
                 if os.path.isfile(stream.temp_path):
                     utils.remove_partial_extension(stream.temp_path)
-                self.validate_file_md5sum(stream)
                 downloaded.append(url)
 
             # Handle file download error, store error to print out later
--- a/gdc_client/parcel/client.py
+++ b/gdc_client/parcel/client.py
@@ -182,15 +182,11 @@ class Client(object):
         return downloaded, errors
 
     def serial_download(self, stream):
-        """Download file to directory serially.
-
-        """
+        """Download file to directory serially."""
         self._download(1, stream)
 
     def parallel_download(self, stream):
-        """Download file to directory in parallel.
-
-        """
+        """Download file to directory in parallel."""
         self._download(self.n_procs, stream)
 
     def _download(self, nprocs, stream):
@@ -223,12 +219,20 @@ class Client(object):
                     if segment is None:
                         return log.debug("Producer returned with no more work")
                     stream.write_segment(segment, producer.q_complete)
+                    # write_segment completed successfully, send sentinal value
+                    # to master process to indicate a task was completed
+                    producer.q_complete.put(None)
                 except Exception as e:
+                    # send sentinal value to master process even though
+                    # write_segment failed to indicate a task is "finished"
+                    producer.q_complete.put(None)
                     if self.debug:
                         raise
                     else:
                         log.error("Download aborted: {0}".format(str(e)))
-                        break
+                        # worker needs to stay alive until final sentinal value
+                        # from master process is received
+                        continue
 
         # Divide work amongst process pool
         pool = [Process(target=download_worker) for i in range(n_procs)]
--- a/gdc_client/parcel/client.py
+++ b/gdc_client/parcel/client.py
@@ -207,20 +207,21 @@ class Client(object):
                 try:
                     segment = producer.q_work.get()
                     if segment is None:
-                        return log.debug("Producer returned with no more work")
+                        log.debug("Producer returned with no more work")
+                        return
                     stream.write_segment(segment, producer.q_complete)
-                    # write_segment completed successfully, send sentinal value
+                    # write_segment completed successfully, send sentinel value
                     # to master process to indicate a task was completed
                     producer.q_complete.put(None)
                 except Exception as e:
-                    # send sentinal value to master process even though
+                    # send sentinel value to master process even though
                     # write_segment failed to indicate a task is "finished"
                     producer.q_complete.put(None)
                     if self.debug:
                         raise
                     else:
                         log.error("Download aborted: {0}".format(str(e)))
-                        # worker needs to stay alive until final sentinal value
+                        # worker needs to stay alive until final sentinel value
                         # from master process is received
                         continue
 
--- a/gdc_client/parcel/client.py
+++ b/gdc_client/parcel/client.py
@@ -110,25 +110,6 @@ class Client(object):
 
         log.debug("Download complete" + rate_info)
 
-    def validate_file_md5sum(self, stream, file_path):
-
-        if not stream.check_file_md5sum:
-            log.debug("checksum validation disabled")
-            return
-
-        log.debug("Validating checksum...")
-
-        if not stream.is_regular_file:
-            raise Exception("Not a regular file")
-
-        if not stream.md5sum:
-            raise Exception(
-                "Cannot validate this file since the server did not provide an md5sum. Use the '--no-file-md5sum' option to ignore this error."
-            )
-
-        if utils.md5sum_whole_file(file_path) != stream.md5sum:
-            raise Exception("File checksum is invalid")
-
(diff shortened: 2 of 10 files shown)


gdc_client/parcel/client.py
```diff
@@ -216,19 +199,30 @@ def _download(self, nprocs, stream):
...
+            return
+
         def download_worker():
             while True:
                 try:
                     segment = producer.q_work.get()
                     if segment is None:
                         return log.debug("Producer returned with no more work")
                     stream.write_segment(segment, producer.q_complete)
+                    # write_segment completed successfully, send sentinal value
```
kulgan:
```suggestion
                    # write_segment completed successfully, send sentinel value
```
?

and can you update L210 to not `return log.debug`. a break statement after the log should do the job

gdc_client/parcel/segment.py
```diff
@@ -320,9 +346,19 @@ def finish_download(self):
...
                 while since_save < self.save_interval:
                     interval = self.q_complete.get()
+                    # Once a process completes a tasks (sucess or failure),
+                    # it will return a sentinal value (None) to main process
+                    # to indicate that a task was completed
+                    if interval is None:
+                        num_tasks_completed += 1
+                        if num_tasks_completed == self.total_tasks:
+                            break
+                        continue
```
kulgan, line 376: Can we use the `data` attribute in the `Interval` class to signal success or failure instead of adding an extra sentinel?

gdc_client/parcel/client.py
jiakf: i can't spell, so i'll change the spelling of sentinel everywhere. 

so L210, it does need to return, as the worker process needs to exit when the main process sends None as a sentinel value to tell it to exit. I can move the debug line above it though and just have a separate return statement under it

gdc_client/parcel/segment.py
jiakf: the producer is already using None as a sentinel value to signal to the worker processes to exit.  Seems functionally the same to have to create a dummy Interval with just a key set in the data attribute and then check that in segment.py to increment the number of tasks completed. if we wanted to take this route, would probably want to update the None being sent as a sentinel to the worker processes to exit with the dummy interval to keep things in sync.

jiakf: and that might not work. if we add a "success" key in the data dict, this would determine whether the small interval chunk of the larger Interval given to a worker process was completed successfully or not, not whether or not the whole Interval or "task" was successful or not.

gdc_client/parcel/utils.py
```diff
@@ -196,6 +188,23 @@ def md5sum_whole_file(fname):
...
+    log.debug("Validating checksum...")
+    if not stream.is_regular_file:
+        raise Exception("Not a regular file")
+
+    if not stream.md5sum:
+        raise Exception(
+            "Cannot validate this file since the server did not provide an md5sum. Use the '--no-file-md5sum' option to ignore this error."
+        )
+    if md5sum_whole_file(file_path) != stream.md5sum:
+        raise Exception("File checksum is invalid")
```
ProfOak: I know this is moved from another section to here but I wonder if this is the opportune time to make these raises a little more specific with a custom Exception like `MD5ValidationError`.

gdc_client/parcel/client.py
```diff
@@ -182,15 +169,11 @@ def download_files(self, urls, *args, **kwargs):
         return downloaded, errors
 
     def serial_download(self, stream):
-        """Download file to directory serially.
-
-        """
+        """Download file to directory serially."""
```
ProfOak, line 172: :clap:

ProfOak (approved):
Just to clarify it sounds like the way you tested this was to run 16 instances of gdc-client, each downloading one >2GB file at a time. It sounds like a hard branch to hit but if this was able to reproduce this issue reliably that is great news.

gdc_client/common/config.py
```diff
@@ -40,7 +40,7 @@ class GDCClientConfigShared(object):
         "n_processes": ConfigParser.getint,
         "retry_amount": ConfigParser.getint,
         "wait_time": ConfigParser.getfloat,
-        "no_segment_md5sum": ConfigParser.getboolean,
+        "no_segment_md5sums": ConfigParser.getboolean,
```
ProfOak, line 43: If we're changing and adding configurable options we should make a note to let user services know.

jiakf: so its a flag that was already there but would never work because of this typo, so with this fix, users should be now getting the functionality they thought they were getting

gdc_client/parcel/utils.py
```diff
@@ -196,6 +190,23 @@ def md5sum_whole_file(fname):
     return hash_md5.hexdigest()
 
 
+def validate_file_md5sum(stream, file_path):
```
BenX00: Could you write a unit test for this? Type hinting and a doc string would be ideal as well.

gdc_client/parcel/segment.py
```diff
@@ -120,83 +122,118 @@ def validate_segment_md5sums(self, path=None):
...
-                ).format(
-                    (
-                        self.download.path
-                        if os.path.isfile(self.download.path)
-                        else self.download.temp_path
-                    ),
-                    self.download.state_path,
+            log.warning("Redownloading {0} corrupt segments.".format(corrupt_segments))
+
+    def recover_intervals(self):
```
BenX00: - Type hint this signature and explain the return
- This method has gotten quite long and complex. Would you be able to break it up into a couple of smaller methods that are more defied in scope to help abstract some of the complexity of this task?
- A lot of debug statements I don't think make sense here. e.g. "State file {0} does not exist. Beginning new download..." Nowere in this code is a new download beginning. As I'm reading it, it happens after the return.
- I'd advocate for a different approach to this method as a whole. The simple boolean return makes reading this really hard cuz its actually doing a lot of different things in the method. Maybe try and have this return the various properties of the object vs setting them and returning a near meaningless bool which you have to explain what it means at every point.

gdc_client/parcel/segment.py
```diff
@@ -120,83 +122,118 @@ def validate_segment_md5sums(self, path=None):
...
+        self.work_pool = IntervalTree([Interval(0, self.download.size)])
+        self.completed = IntervalTree()
+        self.size_complete = 0
+        self.total_tasks = 0
+
+        if not self.recover_intervals():
+            # Recovery failed, treat as new download
+            self.download.setup_file()
+            self.completed = IntervalTree()
+            return
```
BenX00, line 244: Why this return? Was state not loaded here? It reads to me like this is a valid state its just a "new" state.

tests/test_segment_producer.py
```diff
@@ -0,0 +1,197 @@
+import collections
+import logging
+import os
+import pathlib
+import pickle
+import pytest
+import typing
```
BenX00: So was not to long ago directed to this for our style guide and [here](https://google.github.io/styleguide/pyguide.html#31912-imports-for-typing) are its directions for the typing module

gdc_client/parcel/segment.py
```diff
@@ -55,24 +57,24 @@ def __init__(self, download, n_procs):
...
 
     def _setup_work(self):
-        if self.is_complete():
-            log.debug("File already complete.")
-            return
-
         work_size = self.integrate(self.work_pool)
         self.block_size = work_size // self.n_procs
+        self.total_tasks = math.ceil(work_size / self.block_size)
+        log.debug("Total number of tasks: {0}".format(self.total_tasks))
```
BenX00, line 77: this is python 3. Try and use f"" notation.

gdc_client/parcel/segment.py
```diff
@@ -120,83 +122,118 @@ def validate_segment_md5sums(self, path=None):
...
             ), "Bad save state: {0}".format(self.download.state_path)
         except Exception as e:
-            self.completed = IntervalTree()
-            log.error("Unable to resume file state: {0}".format(str(e)))
-        else:
-            self.validate_segment_md5sums(
-                (
+            # An error has occured while loading state file.
+            # Treat as entire file download and recreate temporary file
+            log.error(
```
BenX00, line 167: When we log/catch exceptions its really important to capture the type and the stack trace for the exception. You are eating most of the information which would actually be useful in debugging this code if an unexpected exception was happening in the above block.

tests/test_segment_producer.py
```diff
@@ -0,0 +1,197 @@
...
+
+directories_tuple = collections.namedtuple(
+    "directories", ["base_directory", "data_directory", "state_directory"]
+)
+
+data_info = collections.namedtuple("data_info", ["data", "md5sum"])
+
+
+@pytest.fixture()
+def setup_directories(tmp_path: pathlib.Path):
```
BenX00, line 23: Instead of setting up a bunch of temp files and such make this into an actual unit test and use mocks for the external calls. It will run faster and be easier to arrange things for testing scenarios.

BenX00 (commented):
I haven't looked at everything yet but wanted to get some of what I had out there. Biggest thing I can say is in general more unit tests would be delightful over integration tests.

gdc_client/parcel/utils.py
jiakf: it's not a new function, i just moved the function into the utils module. I can add type hints/docstring and a unit test though for it

gdc_client/parcel/segment.py
jiakf: @ProfOak had brought it up in another comment, I was trying to just keep with the style of the whole codebase. Thought it might be weird to just have a single statement that uses f-string.

gdc_client/parcel/segment.py
jiakf: My changes weren't meant to entirely refactor this method but just get it to a state where it was actually working. The core structure of this method involved logging before a `return` statement. I simply reorganized the logic and fixed a few things that were missing from the method. I'd rather avoid trying to refactor the general structure of this whole method (logging/returns), but can try and break it up into smaller methods. I agree the structure of this method (along with many other things in DTT) are not ideal, but maybe that can be saved for a later ticket

gdc_client/parcel/segment.py
jiakf: had just copied it verbatim from the old code, i think the intention though is that since this log statement will show up for the user, they didn't want the stack trace to be presented to the end user and just the error message to be displayed

gdc_client/parcel/segment.py
jiakf: it's meant to signify that the state is basically "new download". I don't say that this state is invalid, but rather state wasn't loaded.

tests/test_segment_producer.py
jiakf: yep, I had written this months ago and i was not enlightened then/forgot to come back to update this code.

tests/test_segment_producer.py
jiakf: these tests are more-or-less testing the  `load_state` function: they're not true unit tests, but I think when I wrote these tests, i had difficulty coming up with actual unit tests given the code structure of SegmentProducer, and ended up going this route. I think Mocking/patching would get complex here: it might be faster, but I don't think speed is really an issue here.

gdc_client/parcel/segment.py
ProfOak: To add to the confusion loggers have their own string formatting syntax that resembles C's `printf` function. Linters will raise a warning if loggers don't follow this format, too.

```python
log.debug("Total number of tasks: %s", self.total_tasks)
```

It's not exactly the same as `printf` because `%s` can be used with any Object, and it's `__str__()` method gets called to format the string.

tests/test_segment_producer.py
```diff
@@ -0,0 +1,197 @@
...
+    write_state_file(
+        setup_directories.state_directory, "test.txt.parcel", incomplete_data
+    )
+
+
+def write_data_file(directory: pathlib.Path, file_name: str, data: bytes):
+    file_path = directory.joinpath(file_name)
+    with file_path.open("wb") as f:
+        f.write(data)
+
```
ProfOak, line 116: Do any of the consumers of this function clean up the files when they're done?

jiakf: no, from my understanding `tmp_path` creates a new path (which is the base directory where all the created files fall under) for every test run, and pytest will keep the latest 3 `tmp_path` directories (useful for checking file state after a failed test run). So the latest three `tmp_path` directories will remain. If we want to explicitly delete the base directory after each test run we can, its just if you have a failed test for some reason, you won't be able to go inspect the files afterwards (which I ran into).

ProfOak: Oh I see. I misread the `setup_directories` fixture. Type annotations in fixtures trip me up sometimes. I thought `tmp_path` was the parameter and not another fixture.

jiakf: ah, thats a good point. should we not be type hinting fixtures? looking at it now it is kind of confusing

BenX00 (approved):

kulgan (approved):

PR #289, opened 2024-06-20
DEV-2799 fix logging bug

qqiao2024:


kulgan (approved):

.gitlab-ci.yml
```diff
@@ -1,29 +1,12 @@
...
-  before_script:
-    # unshallow the git repo to resolve version with setuptools_scm.
-    - git fetch --unshallow || true
-    - |
-      if [ ${CI_COMMIT_TAG+x} ]; then
-        export TWINE_REPOSITORY_URL=https://nexus.osdc.io/repository/pypi-releases/
-      fi
-  variables:
-    RELEASE_PY_VERSION: python3.6
+      - LANGUAGE_VERSION: [python3.7, python3.8, python3.9]
```
ProfOak, line 12: Thanks for including this range. :+1:

ProfOak (approved):

.gitlab-ci.yml
```diff
@@ -1,29 +1,12 @@
 ---
 include:
   - project: nci-gdc/gitlab-templates
-    ref: 0.6.2
+    ref: master
```
stilesj-uchicago, line 4: Doesn't this need to be a specific version number?

qqiao2024: After @kulgan's update, it should be kinda stable now. So we are using master from now on.

PR #293, opened 2024-07-31
DEV-2882 update deps

qqiao2024:


pyproject.toml
```diff
@@ -40,3 +40,34 @@ skip = [".gitignore", ".dockerignore"]
...
+]
+testpaths = [
+  "tests",
+]
+env = [
+  "LOGSTICK_LOG_FORMAT=raw",
+  "LOGSTICK_LOG_LEVEL=DEBUG",
+  "LOGSTICK_ROOT_LOG_LEVEL=ERROR",
+  "DD_TRACE_GRAPHQL_ENABLED=false"
+]
```
kulgan: need to remove these?

pyproject.toml
```diff
@@ -40,3 +40,34 @@ skip = [".gitignore", ".dockerignore"]
...
+markers = [
+  "config_toggle: for separating out testing different configs"
+]
```
kulgan: Do we have this marker in tests?

pyproject.toml
```diff
@@ -40,3 +40,34 @@ skip = [".gitignore", ".dockerignore"]
...
+minversion = 6.0
+python_files = "test_*.py"
+python_functions = "test_*"
+junit_family = "xunit2"
+junit_logging = "system-out"
+console_output_style = "classic"
+log_level = "ERROR"
+log_cli = "true"
+filterwarnings = ["ignore::DeprecationWarning"]
+addopts = "-rfE --color=yes -p no:ddtrace -p no:ddtrace.pytest_bdd --cov=sheepdog --cov-report term"
```
kulgan: You can remove the filter warnings and also --cov seems to be pointing to the wrong module

kulgan (approved):

pyproject.toml
qqiao2024: removed

pyproject.toml
qqiao2024: removed

pyproject.toml
qqiao2024: removed and fixed.

.github/workflows/ci.yaml
```diff
@@ -5,6 +5,7 @@ on:
       - develop
       - release/**
       - hotfix/**
+      - feat/dev-2882-update-deps
```
wrkhenddher: remove?

setup.cfg
```diff
@@ -25,17 +25,13 @@ package_dir =
...
 install_requires =
-    cryptography
-    jsonschema~=2.6.0
-    lxml~=4.4.2
-    ndg-httpsclient~=0.5.0
-    pyasn1
-    pyOpenSSL
+    jsonschema
+    lxml
     PyYAML>=5.1
```
wrkhenddher: 6.x?

wrkhenddher (approved):

tox.ini
```diff
@@ -6,8 +6,8 @@ allowlist_externals = pytest
 setenv=
   PYTHONHASHSEED=0
 deps=
-  -rdev-requirements.txt
-  .
+  -crequirements.txt
```
stilesj-uchicago, line 9: Why does this move to a constraints (-c) for versions instead of requirements (-r). I feel like this has broke my dev workflow when it was done in a different project when trying to install from the command line and include the `dev` extras.

qqiao2024: 1. Remove dev-requirements.txt remove pins for for develop dependencies. We only need pins for production deps. This remove some extra work for developers. So we only have to use pip-tools to compile requirements.txt, and do not need to compile dev-dependencies anymore.
2. we could use -r requirements.txt here. I used -c requirements.txt here because I think this is faster. If we use -r requirements.txt, pip has to calculate all the dependencies including deps in requirements.txt. By using -c requirements.txt, pip will use them as constraints, and only calculate ".[dev]" dependencies, which suppose to be much faster.
3. When using the requirements.txt file as constrain, the requirements.txt file can not have extra deps for packages. One example I have saw is something like `pyjwt[crypto]==2.8.0`, this will not work if it exists in requirements.txt. I have saw people using `--strip-extras` to remove that.

stilesj-uchicago: I'm not sure if the internet is giving me good information, but I believe requirements and constraints can be used  conjointly like -r requirements.txt -c constraints.txt but they do have different behaviors.

So I guess I have couple of questions:
1. if you try and install the project with a clean python installation will it download and install all the modules you need if only use `-c`? 
2. do we have a reproducible build if we specify constraints but let 'pip install' pull non constrained modules?
3. Is it more appropriate to have a constraints.txt than requirements.txt? Should we be using both?

All of this could be ignorance on my part, but I've seen conflicting information and 'pip install --help' does not give me the advice I need.

qqiao2024: 1. if you try and install the project with a clean python installation will it download and install all the modules you need if only use -c?
The full command is `pip install -c requirements.txt .[dev]`.
`pip install .[dev]` will install all the modules we need for testing (this is tox config file). 
`-c requirements.txt` make sure we are testing the same dependencies we are using in production.

2. do we have a reproducible build if we specify constraints but let 'pip install' pull non constrained modules?
I think for reproducible build, we are building for production. 
This is a tox config file and is for testing. 
And yes, I use the requirements.txt as a constraint file to make sure we are testing the same dependencies in production. 

3. Is it more appropriate to have a constraints.txt than requirements.txt? Should we be using both?
I do not think we should use both. It should be a single source of truth. 
This requirements.txt is also used to install packages in production, it is really a requirements file then. 
It is only used as a constraints file when installing for testing. 
So I think we should keep the name. 

for production build, we are using:
https://github.com/NCI-GDC/gdc-client/blob/develop/bin/package#L74-L75

`pip install --no-deps -r requirements.txt` and `pip install --no-deps .` 
Those will do reproducible build.

Sorry about the confusion, I hope my answers make sense. 
Please let me know if you have more questions or you want me make any changes. 

Thank you!

stilesj-uchicago: Hopefully final question, do you think it is wise to build for development differently than you do for production?

qqiao2024: Dependencies for development has always been different from production, because we need extra packages for development. Like pytest, indexd server, etc.

But, for all the packages we are using in production, we use the same packages when we do testing. That is what the `-c requriements.txt` is for. Which has the same end result as using `-r requirements.txt`. Except use the constraints is faster.

PR #294, opened 2024-09-09
DEV-2987: remove py39 from tox

jiakf:


.gitlab-ci.yml
```diff
@@ -11,4 +11,4 @@ variables:
 tox:
   parallel:
     matrix:
-      - LANGUAGE_VERSION: [python3.8, python3.9]
```
mpsolano, line 14: It looks like the inclusion of python3.9 was not a recent change. The py39 tests did, however, start failing with the previous commit: https://github.com/NCI-GDC/gdc-client/commit/2af71a5d042a78f9f60b15f2638446a1292e5443

Do we know what broke it? If not and if not urgent, should we perhaps take the time to investigate and try to correct?

jiakf: yep, i have more tickets in the detail. I'm assuming we switched gitlab runners to a FIPs runner, and we are getting bit by the md5 issue and the fix for py39 is to set the usedforsecurity=False. I don't think its supported in py38 though, and instead of having to support both versions of python in code: 1) we need to actually run pyupgrade to py39 for gdc-client, which I don't believe we've done and 2) we will need to upgrade to py39 soon and drop py38 support, so the fix will be easier when we make that change.

mpsolano: Sounds good and makes sense to me, thank you for the feedback

mpsolano (approved):
Looks good!

kulgan (commented):
Did you try applying the fix to the line of code that was causing the failure?
The fix will be setting usedforsecurity flag for python39

jiakf:
> Did you try applying the fix to the line of code that was causing the failure? The fix will be setting usedforsecurity flag for python39

i didn't explicitly try it for this PR. I opened up a py38 repl and tried doing md5(usedforsecurity=False) and it threw an exception.

kulgan:
> > Did you try applying the fix to the line of code that was causing the failure? The fix will be setting usedforsecurity flag for python39
> 
> i didn't explicitly try it for this PR. I opened up a py38 repl and tried doing md5(usedforsecurity=False) and it threw an exception.

py38 does not have that flag, so you have to do a conditional check before applying it. You can check a couple other repos where this has been fixed.

jiakf:
> > > Did you try applying the fix to the line of code that was causing the failure? The fix will be setting usedforsecurity flag for python39
> > 
> > 
> > i didn't explicitly try it for this PR. I opened up a py38 repl and tried doing md5(usedforsecurity=False) and it threw an exception.
> 
> py38 does not have that flag, so you have to do a conditional check before applying it. You can check a couple other repos where this has been fixed.

After making the change, it turns out the md5 function is used in code in multiple places. This should be addressed as part of the py39 upgrade for this repo.

kulgan:
> > > > Did you try applying the fix to the line of code that was causing the failure? The fix will be setting usedforsecurity flag for python39
> > > 
> > > 
> > > i didn't explicitly try it for this PR. I opened up a py38 repl and tried doing md5(usedforsecurity=False) and it threw an exception.
> > 
> > 
> > py38 does not have that flag, so you have to do a conditional check before applying it. You can check a couple other repos where this has been fixed.
> 
> After making the change, it turns out the md5 function is used in code in multiple places. This should be addressed as part of the py39 upgrade for this repo.

This [PR](https://github.com/NCI-GDC/gdc-client/pull/295) addresses the md5 usage in fips mode.

kulgan (approved):

PR #298, opened 2026-07-30
DEV-3737 Update python and commit ruff changes

Inarus899:
Mostly ruff changes. Also moved to uv.


src/gdc_client/auth/__init__.py
```diff
@@ -1,2 +1,2 @@
-from . import parser
-from .auth import GDCTokenAuth
+from . import parser as parser
```
stilesj-uchicago: It is unclear to me why this is needed. What happened? It occurs in many __init__.py files.

src/gdc_client/parcel/client.py
```diff
@@ -221,22 +214,22 @@ def download_worker():
...
+        self.start_timer()
 
-        # Start pool
-        for p in pool:
-            p.start()
+        # Divide work amongst process pool
+        from concurrent.futures import ThreadPoolExecutor
 
-        self.start_timer()
+        with ThreadPoolExecutor(max_workers=n_procs) as executor:
```
stilesj-uchicago, line 225: Conversion to a ThreadPoolExecutor is good. I am noting there is no way to stop the processing, but that is no different than the previous implementation.

src/gdc_client/parcel/download_stream.py
```diff
@@ -65,14 +64,12 @@ def setup_file(self):
         self.setup_directories()
         try:
             utils.set_file_length(self.temp_path, self.size)
-        except:
+        except Exception:
             self.log.warning(
```
stilesj-uchicago, line 68: Unclear to my why strip_whitespace is a thing. You can split string across multiple lines by butting them up against each other. Please, rip that out while you're in this codebase.
```suggestion
            self.log.warning((
              f"Unable to set file length. File appears to"
              f" be a {utils.get_file_type(self.path)} file, attempting to proceed"
            ))
```

src/gdc_client/query/versions.py
```diff
@@ -35,13 +35,8 @@ def get_latest_versions(url, uuids, verify=True):
...
             raise HTTPError(
                 (
-                    "The following request {} for ids {} returned with "
-                    "status code: {} and response content: {}"
-                ).format(
-                    versions_url,
-                    chunk,
-                    resp.status_code,
-                    resp.content,
+                    f"The following request {versions_url} for ids {chunk} returned with "
```
stilesj-uchicago, line 38: Here is an example of not using "strip_whitespace"

stilesj-uchicago (changes requested):

src/gdc_client/auth/__init__.py
Inarus899: Ruff complained and suggested this as a solution.

src/gdc_client/query/versions.py
Inarus899: I'm glad with removing strip_whitespace, just wasn't sure if that was going beyond in this case. It is barely used.

stilesj-uchicago (approved):

src/gdc_client/auth/__init__.py
mpsolano: In other PRs where we've implemented ruff, I've seen this handled a little differently and it may be a good idea to handle this consistently from repo to repo. Here's the way I've seen this handled elsewhere:

```
__all__ = ["parser", "GDCTokenAuth"]
```

pyproject.toml
```diff
@@ -61,3 +84,45 @@ norecursedirs = [
...
+  "F", # pyflakes
+  "I", # isort
+  "N", # pep8 naming
+  "RUF", # ruff specific rules
+  "UP", # pyupgrade
+  "W" # pydocstyle warnings
+]
+
+[tool.ruff.lint.per-file-ignores]
+"**/portability.py" = ["F401"]
```
mpsolano: What is the reason for this ignore? I don't think I've seen this done for F401 in other repos and it would be nice to be consistent if possible.

mpsolano (approved):

PR #299, opened 2026-08-10
Test ci matrix upgrade

Inarus899:
This is just for some github, maybe gitlab testing before adding more noise to another pull request.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,6 +38,8 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Install uv
         uses: astral-sh/setup-uv@v9.0.0
+      - name: Install dependencies
+        run: uv sync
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -14,15 +15,15 @@ jobs:
...
-          - macos-12
-          - macos-13
-          - macos-14
-          - ubuntu-20.04
-          - ubuntu-22.04
-          - windows-2019
-          - windows-2022
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
```
Inarus899, line 20: I'm not sure what would be best, but here are the options:
https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories

PR #300, opened 2026-08-11
Feat/dev 3737 matrix update

Inarus899:
This branch continues on the python 3.10-3.14 update and updates the matrix build process for the new versions of python as well as moving things to uv in the process.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,6 +38,8 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Install uv
         uses: astral-sh/setup-uv@v9.0.0
+      - name: Install dependencies
+        run: uv sync
       - name: Run Tests
         run: uv run pytest tests/
       - name: Package
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,9 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v9.0.0
       - name: Install dependencies
-        run: uv sync --no-config
+        run: uv sync
+        env:
+          UV_INDEX_URL: https://pypi.org
       - name: Run Tests
         run: uv run pytest tests/
       - name: Package
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,6 +43,8 @@ jobs:
         env:
           UV_DEFAULT_INDEX: https://pypi.org/simple
       - name: Run Tests
+        env:
+          NO_PROXY: "127.0.0.1,localhost"
         run: uv run pytest tests/
       - name: Package
         run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,6 @@ jobs:
           - macos-14
           - macos-15
           - macos-26
-          - ubuntu-latest
           - ubuntu-22.04
           - ubuntu-24.04
           - windows-2022
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,18 +16,18 @@ jobs:
       matrix:
         os:
           - macos-14
-#          - macos-15
-#          - macos-26
-#          - ubuntu-22.04
-#          - ubuntu-24.04
-#          - windows-2022
-#          - windows-2025
+          - macos-15
+          - macos-26
+          - ubuntu-22.04
+          - ubuntu-24.04
+          - windows-2022
+          - windows-2025
         python:
           - "3.10"
-#          - "3.11"
-#          - "3.12"
-#          - "3.13"
-#          - "3.14"
+          - "3.11"
+          - "3.12"
+          - "3.13"
+          - "3.14"
     env:
       GDC_CLIENT_ZIP: 'gdc-client.zip'
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -14,15 +14,15 @@ jobs:
...
-          - macos-12
-          - macos-13
-          - macos-14
-          - ubuntu-20.04
-          - ubuntu-22.04
-          - windows-2019
-          - windows-2022
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
```
Inarus899: These may or may not be the best options. For a list of options, see:

https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/choose-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories

stilesj-uchicago: Context: we want to be explicit about the versions we support versus those we don't. For instance, we build for python 3.10 as it is the oldest supported version of python. We want to get in any Ubuntu and Linux that is still under LTS (long term support)

.github/workflows/ci.yaml
```diff
@@ -14,15 +15,20 @@ jobs:
     strategy:
       matrix:
         os:
-          - macos-12
-          - macos-13
           - macos-14
-          - ubuntu-20.04
+          - macos-15
+          - macos-26
+          - ubuntu-latest
```
stilesj-uchicago: Maybe remove single non-deterministic one?

stilesj-uchicago (approved):

.github/workflows/ci.yaml
```diff
@@ -5,6 +5,7 @@ on:
       - develop
       - release/**
       - hotfix/**
+      - feat/DEV-3737_Matrix_update
```
BenX00, line 8: We shouldn't check this in.

bin/package
```diff
@@ -24,76 +16,35 @@ with ZipFile('${zip_name}', 'w') as z:
...
+    *) echo "Unsupported OS"; exit 1 ;;
 esac
 
 # setup.py is in previous dir
 cd ..
 
-python -m pip install --upgrade pip
-python -m pip install --no-deps -r requirements.txt
-python -m pip install --no-deps .
+uv sync --locked --extra dev
```
BenX00: Do we need the dev extras? If so these aren't getting packaged are they?

tests/conftest.py
```diff
@@ -141,11 +143,32 @@ def run_mock_server():
 
 @pytest.fixture(scope="class")
 def setup_mock_server() -> None:
-    server = Process(target=run_mock_server)
+    # MacOS does things differently, so get everyone acting the same way
```
mpsolano, line 146: Was this something that changed for mac since py38?

tests/conftest.py
```diff
@@ -141,11 +143,32 @@ def run_mock_server():
...
+        ctx = multiprocessing.get_context("fork")
+        server = ctx.Process(target=run_mock_server)
+    except ValueError:
+        server = multiprocessing.Process(target=run_mock_server)
+
     server.start()
-    time.sleep(5)  # starting with py38, takes longer for process to start on macOS
+
+    # Since py38, a sleep is needed for MacOS. 10 is no longer enough as of py310
+    # Instead of time.sleep(30), this loop could exit faster.
```
mpsolano, line 156: Are you seeing a significant reduction in the wait time for this? This increases complexity a bit and I might recommend just going with the 30 sec sleep if we're only saving a couple seconds.

mpsolano (approved):

tests/conftest.py
Inarus899: I can tell a difference on my machine when running sleep(30) instead of this loop, but how often does a person run the tests? Maybe taking longer and being less complex would be preferred

tests/conftest.py
Inarus899: I can find stuff online saying this change from fork to spawn for just MacOS happened for python 3.8. I don't know how this wasn't an issue for previous builds using python 3.8 on MacOS, but for the build matrix steps, this was causing issues for testing as part of the build process.

.github/workflows/ci.yaml
Inarus899: This line will be removed at the end, but great call out. I tried to make a comment that would block this merge until this line is deleted, but I couldn't find that option quickly enough so I just kept on moving.

.github/workflows/ci.yaml
```diff
@@ -34,17 +39,21 @@ jobs:
...
       - name: Run Tests
-        if: matrix.os != 'windows-latest'
-        run: |
-          pip install --upgrade pip
-          pip install tox
-          tox -e py
+        env:
+          NO_PROXY: "127.0.0.1,localhost"
+          no proxy: "127.0.0.1,localhost"
+        run: uv run --with tox-uv tox
```
BenX00: b/c this isn't a uv project we can just use: `uv run tox`

.github/workflows/ci.yaml
```diff
@@ -34,17 +39,21 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@v9.0.0
+      - name: Install dependencies
+        run: uv sync
+        env:
+          UV_DEFAULT_INDEX: https://pypi.org/simple
```
BenX00: This step shouldn't be required.

Inarus899:
Due to Jira ticket hooks and slack, I created a new pull request (just tried to change the name really) and it can be found here:
https://github.com/NCI-GDC/gdc-client/pull/301

PR #303, opened 2026-09-03
Feat/dev 4238 context managers

Inarus899:
Add context managers to reduce the number of connections that never close.

### DIFF
--- a/src/gdc_client/client/client.py
+++ b/src/gdc_client/client/client.py
@@ -1,4 +1,4 @@
-from contextlib import closing, contextmanager
+from contextlib import contextmanager
 
 import requests
 
@@ -29,17 +29,19 @@ class GDCClient:
             "User-Agent": agent,
         }
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.session.close()
+
     @contextmanager
     def request(self, verb, path, **kwargs):
         """Make a request to the GDC API."""
-        res = self.session.request(
-            verb,
-            f"https://{self.host}:{self.port}{path}",
-            auth=auth.GDCTokenAuth(self.token),
-            **kwargs,
-        )
+        url = f"https://{self.host}:{self.port}{path}"
+        auth_handler = auth.GDCTokenAuth(self.token)
 
-        with closing(res):
+        with self.session.request(verb, url, auth=auth_handler, **kwargs) as res:
             yield res
 
     def get(self, path, **kwargs):
--- a/src/gdc_client/client/client.py
+++ b/src/gdc_client/client/client.py
@@ -1,4 +1,7 @@
-from contextlib import contextmanager
+import collections.abc
+import contextlib
+import types
+import typing
 
 import requests
 
@@ -32,17 +35,24 @@ class GDCClient:
     def __enter__(self):
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: types.TracebackType | None,
+    ) -> None:
         self.session.close()
 
-    @contextmanager
-    def request(self, verb, path, **kwargs):
+    @contextlib.contextmanager
+    def request(
+        self, verb: str, path: str, **kwargs: typing.Any
+    ) -> collections.abc.Iterator[requests.Response]:
         """Make a request to the GDC API."""
         url = f"https://{self.host}:{self.port}{path}"
         auth_handler = auth.GDCTokenAuth(self.token)
 
-        with self.session.request(verb, url, auth=auth_handler, **kwargs) as res:
-            yield res
+        with self.session.request(verb, url, auth=auth_handler, **kwargs) as response:
+            yield response
 
     def get(self, path, **kwargs):
         return self.request("GET", path, **kwargs)
--- a/src/gdc_client/download/client.py
+++ b/src/gdc_client/download/client.py
@@ -1,9 +1,9 @@
+import contextlib
 import logging
 import os
 import re
 import tarfile
 import time
-from contextlib import contextmanager
 from io import BytesIO
 from urllib import parse as urlparse
 
@@ -114,17 +114,20 @@ class GDCHTTPDownloadClient(HTTPClient):
             ann_ids = {"ids": annotations}
 
             # NOTE: Force compression
-            with self._post(path="data?compress", json=ann_ids) as r:
-                r.raise_for_status()
-                tar = tarfile.open(mode="r:gz", fileobj=BytesIO(r.content))
-                if self.annotation_name in tar.getnames():
-                    member = tar.getmember(self.annotation_name)
-                    ann = tar.extractfile(member).read()
-                    path = os.path.join(directory, self.annotation_name)
-                    with open(path, "wb") as f:
-                        f.write(ann)
+            with self._post(path="data?compress", json=ann_ids) as response:
+                response.raise_for_status()
+                with (
+                    BytesIO(response.content) as buffer,
+                    tarfile.open(mode="r:gz", fileobj=buffer) as tar,
+                ):
+                    if self.annotation_name in tar.getnames():
+                        member = tar.getmember(self.annotation_name)
+                        ann = tar.extractfile(member).read()
+                        path = os.path.join(directory, self.annotation_name)
+                        with open(path, "wb") as f:
+                            f.write(ann)
 
-                    log.debug(f"Wrote annotations to {path}.")
+                        log.debug(f"Wrote annotations to {path}.")
 
     def _untar_file(self, tarfile_name):
         # type: (str) -> list[str]
@@ -163,7 +166,7 @@ class GDCHTTPDownloadClient(HTTPClient):
 
         return errors
 
-    @contextmanager
+    @contextlib.contextmanager
     def _post(self, path, headers=None, json=None, stream=True):
         # type: (str, dict[str,str], dict[str,object], bool) -> requests.models.Response
         """custom post request that will query both active and legacy api
@@ -182,19 +185,19 @@ class GDCHTTPDownloadClient(HTTPClient):
                 verify=self.verify,
                 json=json or {},
                 headers=headers or {},
-            ) as r:
-                if r.status_code not in [200, 203]:
-                    # try legacy if active doesn't return OK
-                    with requests.post(
-                        legacy,
-                        stream=stream,
-                        verify=self.verify,
-                        json=json or {},
-                        headers=headers or {},
-                    ) as r_legacy:
-                        yield r_legacy
-                else:
-                    yield r
+            ) as response:
+                if response.status_code in [200, 203]:
+                    yield response
+                    return
+                # try legacy if active doesn't return OK
+                with requests.post(
+                    legacy,
+                    stream=stream,
+                    verify=self.verify,
+                    json=json or {},
+                    headers=headers or {},
+                ) as response:
+                    yield response
 
         except Exception as e:
             log.error(e)
@@ -214,28 +217,28 @@ class GDCHTTPDownloadClient(HTTPClient):
         # POST request avoids the MAX LEN character limit for URLs
         params = ("tarfile",)
         path = build_url("data", *params)
-        with self._post(path=path, headers=headers, json=ids) as r:
-            if r.status_code == requests.codes.bad:
+        with self._post(path=path, headers=headers, json=ids) as response:
+            if response.status_code == requests.codes.bad:
                 log.error("Unable to connect to the API")
                 log.error(f"Is this the correct URL? {self.base_uri}")
 
-            elif r.status_code == requests.codes.forbidden:
+            elif response.status_code == requests.codes.forbidden:
                 # since the files are grouped by access control, that means
                 # a group is entirely controlled or open access.
                 # If it fails to download because you don't have access then
                 # don't bother trying again
-                log.error(r.text)
+                log.error(response.text)
                 return "", []
 
-            if r.status_code not in [200, 203]:
-                log.warning(f"[{r.status_code}] Unable to download group")
+            if response.status_code not in [200, 203]:
+                log.warning(f"[{response.status_code}] Unable to download group")
                 errors.append(ids["ids"])
                 return "", errors
 
             # {'content-disposition': 'filename=the_actual_filename.tar'}
-            content_filename = r.headers.get("content-disposition") or r.headers.get(
-                "Content-Disposition"
-            )
+            content_filename = response.headers.get(
+                "content-disposition"
+            ) or response.headers.get("Content-Disposition")
 
             if content_filename:
                 tarfile_name = os.path.join(
@@ -246,7 +249,7 @@ class GDCHTTPDownloadClient(HTTPClient):
                 tarfile_name = time.strftime("gdc-client-%Y%m%d-%H%M%S.tar")
 
             with open(tarfile_name, "wb") as f:
-                for chunk in r:
+                for chunk in response:
                     f.write(chunk)
 
         return tarfile_name, errors
@@ -296,8 +299,6 @@ class GDCHTTPDownloadClient(HTTPClient):
         return errors, successful_count
 
     def parallel_download(self, stream):
-        # gdc-client calls parcel's parallel_download,
-        # which is where most of the downloading takes place
         file_id = stream.url.split("/")[-1]
         super().parallel_download(stream)
 
--- a/src/gdc_client/download/client.py
+++ b/src/gdc_client/download/client.py
@@ -3,6 +3,7 @@ import os
 import re
 import tarfile
 import time
+from contextlib import contextmanager
 from io import BytesIO
 from urllib import parse as urlparse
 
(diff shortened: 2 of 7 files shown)


src/gdc_client/parcel/download_stream.py
```diff
@@ -40,6 +40,13 @@ def init(self):
         self.initialized = True
         return self
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if hasattr(self, "_session"):
+            self._session.close()
```
stilesj-uchicago, line 50: No _session is created or defined in this class.

stilesj-uchicago (dismissed):

src/gdc_client/client/client.py
```diff
@@ -29,17 +29,19 @@ def __init__(self, host=GDC_API_HOST, port=GDC_API_PORT, token=None):
...
-            verb,
-            f"https://{self.host}:{self.port}{path}",
-            auth=auth.GDCTokenAuth(self.token),
-            **kwargs,
-        )
+        url = f"https://{self.host}:{self.port}{path}"
+        auth_handler = auth.GDCTokenAuth(self.token)
 
-        with closing(res):
+        with self.session.request(verb, url, auth=auth_handler, **kwargs) as res:
```
BenX00: Try and use full words vs abbreviations; [see](https://google.github.io/styleguide/pyguide.html#316-naming).
```suggestion
        with self.session.request(verb, url, auth=auth_handler, **kwargs) as response:
```

src/gdc_client/client/client.py
```diff
@@ -29,17 +29,19 @@ def __init__(self, host=GDC_API_HOST, port=GDC_API_PORT, token=None):
...
         }
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.session.close()
+
     @contextmanager
     def request(self, verb, path, **kwargs):
```
BenX00: Since you are here always nice to type hint the method/function you are in.
```suggestion
    def request(self, verb: str, path: str, **kwargs: Any) -> Iterator[requests.Response]:
```

src/gdc_client/client/client.py
```diff
@@ -1,4 +1,4 @@
-from contextlib import closing, contextmanager
+from contextlib import contextmanager
```
BenX00: Per the style guide, always import the module. This applies for all imports in this PR

src/gdc_client/download/client.py
```diff
@@ -162,41 +163,42 @@ def _md5_members(self, members):
...
+                    with requests.post(
+                        legacy,
+                        stream=stream,
+                        verify=self.verify,
+                        json=json or {},
+                        headers=headers or {},
+                    ) as r_legacy:
+                        yield r_legacy
+                else:
+                    yield r
```
BenX00: Be mindful of your contexts. This could ultimately keep two connections alive for no reason.

Some useful reading:
- [StopIteration](https://docs.python.org/3/library/exceptions.html#StopIteration)
- [Yield](https://docs.python.org/3/reference/expressions.html#yield-expressions)
- [Generator-Iterators](https://docs.python.org/3/reference/expressions.html#generator-iterator-methods)
```suggestion
            with requests.post(
                active,
                stream=stream,
                verify=self.verify,
                json=json or {},
                headers=headers or {},
            ) as response:
                if r.status_code in [200, 203]:
                    yield response
                    # Return can be used in a generator to exit early.
                    # Effectively this raises StopIteration
                    return  
            # Close previous connection before trying the next one.
                    
            # try legacy if active doesn't return OK
            with requests.post(
                legacy,
                stream=stream,
                verify=self.verify,
                json=json or {},
                headers=headers or {},
            ) as response:
                yield response
```

src/gdc_client/download/client.py
```diff
@@ -212,43 +214,40 @@ def _download_tarfile(self, small_files):
...
-        # {'content-disposition': 'filename=the_actual_filename.tar'}
-        content_filename = r.headers.get("content-disposition") or r.headers.get(
-            "Content-Disposition"
-        )
-
-        if content_filename:
-            tarfile_name = os.path.join(
-                self.base_directory,
-                content_filename.split("=")[1],
+        with self._post(path=path, headers=headers, json=ids) as r:
```
BenX00:
```suggestion
        with self._post(path=path, headers=headers, json=ids) as response:
```

src/gdc_client/client/client.py
```diff
@@ -29,17 +29,19 @@ def __init__(self, host=GDC_API_HOST, port=GDC_API_PORT, token=None):
             "User-Agent": agent,
         }
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
BenX00: Just because a lot of this code base isn't typed doesn't mean we can make that problem worse. Please ALWAYS annotate your code. If you need help with the proper types feel free to ask.

src/gdc_client/download/client.py
```diff
@@ -113,17 +114,17 @@ def download_annotations(self, file_id):
...
-                member = tar.getmember(self.annotation_name)
-                ann = tar.extractfile(member).read()
-                path = os.path.join(directory, self.annotation_name)
-                with open(path, "wb") as f:
-                    f.write(ann)
-
-                log.debug(f"Wrote annotations to {path}.")
+            with self._post(path="data?compress", json=ann_ids) as r:
+                r.raise_for_status()
+                tar = tarfile.open(mode="r:gz", fileobj=BytesIO(r.content))
```
BenX00: Oh yay some IO. This should also be in a with block.
```suggestion
                with BytesIO(response.content) as buffer, tarfile.open(mode="r:gz", fileobj=buffer) as tar:
```

src/gdc_client/parcel/client.py
```diff
@@ -129,29 +129,32 @@ def download_files(self, urls, *args, **kwargs):
...
-                    log.exception(e)
-                    raise
-
-            finally:
-                utils.print_closing_header(url)
+            with DownloadStream(url, self.directory, self.token) as stream:
+                # Download file
+                try:
+                    # validate temporary file before renaming to permanent file location
+                    self.parallel_download(stream)
```
BenX00, line 136: If I'm reading correctly, this stream is used across multiple threads? Which is a problem b/c the session object is not thread safe if I remember correctly.

src/gdc_client/parcel/client.py
```diff
@@ -238,18 +241,15 @@ def _standard_tcp_download(self, stream):
...
-                stream.setup_directories()
-                with open(stream.path, "wb") as f:
-                    for chunk in r:
-                        f.write(chunk)
-
-            else:
-                raise Exception(f"[{r.status_code}] Unable to download url {stream.url}")
-
-            r.close()
+            with requests.get(stream.url, stream=True, verify=self.verify) as r:
```
BenX00:
```suggestion
            with requests.get(stream.url, stream=True, verify=self.verify) as response:
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -161,19 +168,17 @@ def request(self, headers=None, verify=True, close=False, max_retries=16):
                 stream=True,
                 timeout=max_timeout,
             )
+            r.raise_for_status()
+
+            if close:
+                r.close()
+            return r
+
```
BenX00, line 178: This leaves the stream open 👎

BenX00 (changes requested):

src/gdc_client/parcel/client.py
BenX00: In the long run, we could consider switching to [httpx](https://www.python-httpx.org/) which is [thread-safe](https://github.com/encode/httpx/discussions/1633)

src/gdc_client/parcel/download_stream.py
Inarus899: I am open for improvements, but I got to this spot because wrapping this code up using "with" causes a test to fail, one designed to retry downloads. The reason I think this is ok is because this function is only called in this repo twice, and both of those are using context managers.

src/gdc_client/parcel/download_stream.py
Inarus899: I am now creating _session as part of the init process, but we do have a test that shows that request() in this class can be reached when _session is not set, so there is a work around there. I am going to do further research, just trying to get this out ahead of the weekend.

stilesj-uchicago: There is definitely going to be some weird state here:
- class initialized but no context manager called - self._session is None
- context manager is used and exited - self._session is closed which would error if  `request` is called again.

@BenX00 , do you have any insights on how to detangle this? I was thinking that the _enter_ method might return the class Session.

src/gdc_client/download/client.py
```diff
@@ -162,41 +166,42 @@ def _md5_members(self, members):
...
-                r = requests.post(
+                with requests.post(
                     legacy,
                     stream=stream,
                     verify=self.verify,
                     json=json or {},
                     headers=headers or {},
-                )
+                ) as response:
+                    yield response
```
BenX00, line 200: Be very conscious of when a context begins and ends. No compiler/hinting can help you here. You'll want to close the first request if it fails before trying to make another.
```suggestion
            with requests.post(
                legacy,
                stream=stream,
                verify=self.verify,
                json=json or {},
                headers=headers or {},
            ) as response:
                yield response
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -18,6 +22,23 @@
 from gdc_client.parcel.defaults import max_timeout
 
 
+class SessionCache:
```
BenX00, line 25: Would you add some docstrings for the class and the public methods?

src/gdc_client/parcel/download_stream.py
```diff
@@ -18,6 +22,23 @@
...
+    def __init__(self) -> None:
+        self._lock = threading.Lock()
+        self._context = contextlib.ExitStack()
+
+    def close(self, *args, **kwargs) -> None:
```
BenX00:
```suggestion
    def close(self) -> None:
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -33,13 +54,28 @@ def __init__(self, url, directory, token=None):
...
         self.get_information()
         self.print_download_information()
         self.initialized = True
         return self
 
+    def __enter__(self):
+        # Using threading.local keeps each session in a single thread
+        self._session_storage = threading.local()
+        self._session_storage.session = None
+        return self
```
BenX00: These lines should be removed.

src/gdc_client/parcel/download_stream.py
```diff
@@ -133,7 +169,7 @@ def header(self, start=None, end=None):
             header["host"] = host
         return header
 
-    def request(self, headers=None, verify=True, close=False, max_retries=16):
+    def request(self, headers: dict[str, Any], verify: bool = True, max_retries: int = 16):
```
BenX00:
```suggestion
    @contextlib.contextmanager
    def request(self, headers: dict[str, Any], verify: bool = True, max_retries: int = 16) -> Iterator[requests.Request]:
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -18,6 +22,23 @@
...
+    def close(self, *args, **kwargs) -> None:
+        self._context.close()
+
+    def get_session(self) -> requests.Session:
+        with self._lock:
+            if not hasattr(threading.local(), "session"):
+                local_thread = threading.local()
+                local_thread.session = self._context.enter_context(requests.Session())
+
+        return local_thread.session
```
BenX00: Static type checking says.... ERROR 😁 What happens in the case `session` is set? What is `local_thread` set to?

src/gdc_client/parcel/download_stream.py
BenX00:
```suggestion
        return self
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -33,13 +54,28 @@ def __init__(self, url, directory, token=None):
...
+        self._session_storage.session = None
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: types.TracebackType | None,
+    ) -> None:
+        self._session_storage.session = None
```
BenX00:
```suggestion
        self._session_cache.close()
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -18,6 +22,41 @@
...
+        If a session does not already exist for the current thread, a new
+        request.Session instance is created and added to the ExitStack context
+        to manage closing out connections as expected, including during error
+        handling.
+
+        Returns:
+            requests.Session: A session object isolated to the local thread.
+        """
+        with self._lock:
+            if not hasattr(threading.local(), "session"):
```
BenX00:
```suggestion
            if not hasattr(self._local, "session"):
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -33,13 +72,26 @@ def __init__(self, url, directory, token=None):
...
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: types.TracebackType | None,
+    ) -> None:
+        if hasattr(self._session_cache._local, "session"):
+            self._session_cache._local.session = None
```
BenX00: We want to make sure to close all the sessions.
```suggestion
        self._session_cache.close()
```

src/gdc_client/parcel/download_stream.py
```diff
@@ -184,34 +236,34 @@ def get_information(self):
...
 
-        self.md5sum = None
-        if self.check_file_md5sum:
-            self.md5sum = r.headers.get("content-md5", "")
+            self.md5sum = None
+            if self.check_file_md5sum:
+                self.md5sum = response.headers.get("content-md5", "")
 
-        return self.name, self.size
+            return self.name, self.size
```
BenX00: I tested this (so this is my bad cuz I hadn't yesterday) you can unnest everything here. You can access the response headers safely outside of the `with` block.

src/gdc_client/query/index.py
```diff
@@ -54,20 +54,17 @@ def _get_hits(self, url, metadata_query):
...
+            if json_response.get("data") is None or json_response["data"].get("hits") is None:
+                return []
 
-        r.close()
-
-        if json_response.get("data") is None or json_response["data"].get("hits") is None:
-            return []
-
-        return json_response["data"]["hits"]
+            return json_response["data"]["hits"]
```
BenX00: This all can be simplified to:
```python
            return json_response.get("data", {}).get("hits", [])
```

BenX00 (approved):
Just some minor things. Tho the self._local check and the closing of the cache are critical.

src/gdc_client/download/client.py
```diff
@@ -82,17 +83,16 @@ def download_related_files(self, file_id):
             for related_file in related_files:
                 log.debug(f"related file {related_file}")
                 related_file_url = urlparse.urljoin(self.data_uri, related_file)
-                stream = DownloadStream(related_file_url, directory, self.token)
+                with DownloadStream(related_file_url, directory, self.token) as stream:
+                    # TODO: un-set this when parcel is moved to dtt
```
mpsolano, line 87: Is this TODO something planned to be done before this PR is merged or is it future work? If it's future work, is there a ticket associated with that?

mpsolano (approved):

src/gdc_client/client/client.py
```diff
@@ -29,18 +32,27 @@ def __init__(self, host=GDC_API_HOST, port=GDC_API_PORT, token=None):
...
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: types.TracebackType | None,
+    ) -> None:
+        self.session.close()
+
+    @contextlib.contextmanager
+    def request(
+        self, verb: str, path: str, **kwargs: typing.Any
```
stilesj-uchicago: There is an exception `2.2.4.1` for using `typing.` in the [style guide](https://google.github.io/styleguide/pyguide.html)

src/gdc_client/download/client.py
```diff
@@ -162,41 +165,42 @@ def _md5_members(self, members):
...
-                    stream=stream,
-                    verify=self.verify,
-                    json=json or {},
-                    headers=headers or {},
-                )
+            ) as response:
+                if response.status_code in [200, 203]:
+                    yield response
+                    return
+            # try legacy if active doesn't return OK
```
stilesj-uchicago, line 191: I'm not sure if we need this any more. I peeked at the gdcapi and it looks like requests for legacy files always return an error now. No code change needed though.

src/gdc_client/download/client.py
```diff
@@ -212,43 +216,40 @@ def _download_tarfile(self, small_files):
...
-            )
-        else:
-            tarfile_name = time.strftime("gdc-client-%Y%m%d-%H%M%S.tar")
-
-        with open(tarfile_name, "wb") as f:
-            for chunk in r:
-                f.write(chunk)
+        with self._post(path=path, headers=headers, json=ids) as response:
+            if response.status_code == requests.codes.bad:
+                log.error("Unable to connect to the API")
```
stilesj-uchicago: `requests.codes.bad` maps to 400. If a 400 was returned it could connect to the API but the client made a bad request. Please, fix the error message.

Further, this does not `return "", []` like the `requests.codes.forbidden` check does. Is that correct? A 400 results in further processing?

src/gdc_client/parcel/download_stream.py
```diff
@@ -184,10 +235,10 @@ def get_information(self):
         """
 
         headers = self.header()
-        r = self.request(headers, close=True)
-        self.log.debug("Request responded")
+        with self.request(headers) as response:
```
stilesj-uchicago, line 238: I notice that this assumes a response and does not handle an error case. May a `response.raise_for_status()` is appropriate?

src/gdc_client/parcel/download_stream.py
```diff
@@ -235,27 +286,26 @@ def write_segment(self, segment, q_complete, retries=5):
...
-                utils.write_offset(self.temp_path, chunk, offset)
-                if self.check_segment_md5sums:
-                    iv_data = {"md5sum": utils.md5sum(chunk)}
-                else:
-                    iv_data = None
-                complete_segment = Interval(offset, offset + len(chunk), iv_data)
-                q_complete.put(complete_segment)
-
-                written += len(chunk)
+            with self.request(self.header(start, end)) as response:
```
stilesj-uchicago, line 290: assumes successful response.

src/gdc_client/download/client.py
Inarus899: There are a lot of TODO in this code, and none of them were from me.

stilesj-uchicago (approved):

PR #304, opened 2026-09-17
Feat/dev 4281 macos 26 work

Inarus899:
A problem was found for MacOS 26 python 3.14, and the fix is a flag for pyinstaller


mpsolano (approved):

bin/package
```diff
@@ -29,11 +29,12 @@ done
 
 DATE=${DATE-} # if date is not passed in, default DATE variable to empty string
 BINARY_NAME="gdc-client"
+OSX_FLAGS=""
 
 case "$(uname -s)" in
     *NT*) TARGET_ENVIRONMENT="Windows"; BINARY_NAME="gdc-client.exe" ;;
     *Linux*) TARGET_ENVIRONMENT="Ubuntu" ;;
-    *Darwin*) TARGET_ENVIRONMENT="OSX" ;;
+    *Darwin*) TARGET_ENVIRONMENT="OSX"; OSX_FLAGS="--codesign-identity -" ;;
```
stilesj-uchicago, line 37: What is the trailing `-` before the last quote? Is that intentional?

Inarus899: It is intentional. The hyphen is used to make macos-26 use ad-hoc code signing, the reason why there was different outcomes for macos-15 and macos-26 in testing.

stilesj-uchicago (approved):

